PR #22, opened 2019-12-11
feat: migrate BQ Java samples from java-docs-samples to client lib

stephaniewang526:
Fixes b/146052864

### DIFF
--- a/samples/pom.xml
+++ b/samples/pom.xml
@@ -44,14 +44,19 @@
     </dependency>
     <!-- [END bigquery_java_dependencies] -->
     <dependency>
-      <groupId>commons-cli</groupId>
-      <artifactId>commons-cli</artifactId>
-      <version>1.4</version>
+      <groupId>com.google.api</groupId>
+      <artifactId>gax</artifactId>
+      <version>1.49.1</version>
     </dependency>
     <dependency>
-      <groupId>joda-time</groupId>
-      <artifactId>joda-time</artifactId>
-      <version>2.10.5</version>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-core</artifactId>
+      <version>1.91.3</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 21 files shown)


samples/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.11</version>
+  </parent>
+
+  <properties>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
chingor13: If we're compiling for Java 8, we need to skip this entire module for the Java 7 CI. We can do that with a profile in the root project where we conditionally change the list of modules.

samples/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <!-- [START bigquery_java_dependencies] -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigquery</artifactId>
+      <version>1.100.0</version>
```
chingor13: This is an interesting problem to solve - we want to test against HEAD (usually the `-SNAPSHOT` version), but for the sample, we want the released version.

This can be solved in a separate PR.

samples/pom.xml
stephaniewang526: Updated PR

samples/pom.xml
stephaniewang526: I will keep the released version for now since this is the safest thing. We will need to look into incorporating release process into samples release.

BenWhitehead:
We also need to disable release for the sample module. @chingor13 How do we do that with the current setup of things.

PR #32, opened 2019-12-13
feat: skip release/staging for sample artifacts

### LINKED ISSUE #23: Skip release/staging for sample artifacts
Figure out how to structure the repo so that we skip deploy/staging for the samples artifact.

stephaniewang526:
Fixes #23

### DIFF
--- a/samples/pom.xml
+++ b/samples/pom.xml
@@ -74,4 +74,25 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <version>2.8.2</version>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.sonatype.plugins</groupId>
+        <artifactId>nexus-staging-maven-plugin</artifactId>
+        <version>1.6.8</version>
+        <configuration>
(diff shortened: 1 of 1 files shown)


samples/pom.xml
```diff
@@ -74,4 +74,25 @@
...
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.sonatype.plugins</groupId>
+        <artifactId>nexus-staging-maven-plugin</artifactId>
+        <version>1.6.8</version>
+        <configuration>
+          <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
```
BenWhitehead: It looks like we may have a potential ordering issue using this property due to the way the staging plugin defers to the final module loaded in the project.

From the docs:

> `skipNexusStagingDeployMojo`
> 
> Set to false by default, this flag will cause to skip any execution of the deploy goal of the plugin when set to true similar to maven.deploy.skip. In multi-module builds the staging of all components is performed in the last module based on the reactor order. If this property is set to true in the module, all staging will be skipped. You have to ensure that this property evaluates as true in the last module of the reactor. If necessary, you can add a dummy module.

If this is the case, we may need to change how the samples profile is activated.

@chingor13 What are your thoughts on this?

BenWhitehead:
@chingor13 Rather than skipping the samples by default, we may be able to standardize on what the profile name is for samples in synthtool and then explicitly disable that profile[1] when performing a release so that it isn't picked up.

[1] http://maven.apache.org/guides/introduction/introduction-to-profiles.html#Deactivating_a_profile

PR #38, opened 2019-12-16
feat: use libraries-bom for samples

stephaniewang526:


samples/pom.xml
```diff
@@ -36,12 +36,25 @@
...
 
+  <!-- [START bigquery_java_dependencies] -->
+  <!--  Using libraries-bom to manage versions.
+  See https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/The-Google-Cloud-Platform-Libraries-BOM -->
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>3.0.0</version>
```
BenWhitehead: `3.2.0` was just released, lets use that instead of `3.0.0`.

stephaniewang526: will leave as 3.0.0

PR #44, opened 2019-12-17
feat: add samples

stephaniewang526:
Fixes b/146347707, b/146452106

PR #52, opened 2019-12-19
feat: update samples

stephaniewang526:
follow up PR of #44


samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,29 +1,84 @@
...
+      if(job != null && job.getStatus().getError() == null)
+        System.out.println("Table extraction job completed successfully. Check in GCS bucket for the CSV file.");
+      else
+        System.out.println("Table extraction job failed");
     } catch (BigQueryException e) {
       System.out.println("Table extraction job was interrupted. \n" + e.toString());
     }
   }
+
+  private static Table createTableHelper(String datasetName, String tableName, Schema schema) {
```
kurtisvg: What is the point of this helper? 

If the goal of the snippet is to show the user how to extract data from a table, we should assume they already have a table created and show them how to initialize the model of the resource in their current environment (probably from  a table name?)

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,29 +1,84 @@
...
 
   // Exports my-dataset-name:my_table to gcs://my-bucket/my-file as raw CSV
   public static void extractTableToJSON(Table table, String format, String gcsFileName) {
-
     try {
-      table.extract(format, gcsFileName);
-      System.out.println("Table extraction job completed successfully");
+      Job job = table.extract(format, gcsFileName);
+      if(job != null && job.getStatus().getError() == null)
+        System.out.println("Table extraction job completed successfully. Check in GCS bucket for the CSV file.");
```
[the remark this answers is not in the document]
stephaniewang526: shared-config is already our parent pom

https://github.com/googleapis/java-bigquery/blob/3f213c29ebc437ec79c04553db8b2675efb3aa54/pom.xml#L16

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,29 +1,84 @@
...
 public class ExtractTableToJSON {
 
   public static void runExtractTableToJSON() {
     // TODO(developer): Replace these variables before running the sample.
-    Table table = null;
+    String datasetName = "MY_DATASET_NAME";
     String format = "CSV";
     String bucketName = "my-bucket";
-    String gcsFileName = "gs://" + bucketName + "/extractTest.csv";
+    String gcsFileName = "gs://" + bucketName + "/path/to/file";
```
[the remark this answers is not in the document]
stephaniewang526: Python sample (our canonical) is naming it as `destination_uri` -- I will be doing the same

https://github.com/googleapis/google-cloud-python/blob/371c4f270b9a8c3e1c3b87e6edfdda1a61ae5e89/bigquery/docs/snippets.py#L1241

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,29 +1,84 @@
...
+    // once, and can be reused for multiple requests.
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+    TableId tableId = TableId.of(datasetName, tableName);
+    TableDefinition tableDefinition = StandardTableDefinition.of(schema);
+    TableInfo tableInfo = TableInfo.newBuilder(tableId, tableDefinition).build();
+
+    try {
+      Table table = bigquery.create(tableInfo);
+      return table;
```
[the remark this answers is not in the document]
stephaniewang526: Getting rid of the createTabler Helper function

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,29 +1,84 @@
...
+    } catch (BigQueryException e) {
+      System.out.println("Table was not created. \n" + e.toString());
+      return null;
```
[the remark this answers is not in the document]
stephaniewang526: Getting rid of the createTabler Helper function

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
stephaniewang526: Getting rid of the createTabler Helper function

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,29 +1,84 @@
...
+import com.google.cloud.bigquery.TableId;
+import com.google.cloud.bigquery.TableInfo;
 
 public class ExtractTableToJSON {
 
   public static void runExtractTableToJSON() {
     // TODO(developer): Replace these variables before running the sample.
-    Table table = null;
+    String datasetName = "MY_DATASET_NAME";
     String format = "CSV";
```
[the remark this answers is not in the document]
stephaniewang526: Adding link: https://cloud.google.com/bigquery/docs/exporting-data#export_formats_and_compression_types

samples/src/main/java/com/example/bigquery/CreateTable.java
```diff
@@ -16,11 +32,11 @@
 
   public static void runCreateTable() {
     // TODO(developer): Replace these variables before running the sample.
-    String datasetName = "my-dataset-name";
-    String tableName = "my_table_name";
+    String datasetName = "MY_DATASET_NAME";
+    String tableName = "MY_TABLE_NAME";
     Schema schema =
```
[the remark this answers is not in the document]
stephaniewang526: Getting rid of creatTable helper function so schema will no longer be used

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,26 +1,61 @@
...
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests.
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+    TableId tableId = TableId.of(projectId, datasetName, tableName);
+    Table table = bigquery.getTable(tableId);
 
+    // For more information on export format available see:
+    // https://cloud.google.com/bigquery/docs/exporting-data#export_formats_and_compression_types
+    Job job = table.extract("CSV", destinationUri);
```
kurtisvg: `Job` seems to indicate that this is a Long Running Operation (LRO) - do you need to do something to wait for the Job to complete before you get results?

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,26 +1,61 @@
...
     try {
-      table.extract(format, gcsFileName);
-      System.out.println("Table extraction job completed successfully");
+      if (job != null && job.getStatus().getError() == null)
```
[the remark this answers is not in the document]
kurtisvg: Also - under what circumstances would job be `null`? Does `null` being returned mean something significant?

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,26 +1,61 @@
...
+        System.out.println(
+            "Table extraction job completed successfully. Check in GCS bucket for the CSV file.");
+      else System.out.println("Table extraction job failed");
```
kurtisvg: nit: Generally, it's easier to read/understand control flow logic if you use this pattern:
```
if (errorCondition){
   // do error logic
}
// do success logic
```

If the error is a terminal condition, just return after logging:
```
if (errorCondition){
   // do error logic
  return;
}
// do success logic
```

This is easier to understand for two reasons:
- it reinforces `success` as the normal flow
- it resolves one branch of logic (if error) before tackling the next branch (if success)

So consider something like the following:
```java
if (job == null || job.getStatus().getError() != null) {
   System.out.println("BigQuery was unable to extract due to an error: \n" + job.getStatus().getError());
   return;
}
System.out.println("Table export successful. Check in GCS bucket for the CSV file.");
```

samples/src/test/java/com/example/bigquery/ExtractTableToJSONIT.java
```diff
@@ -38,39 +26,9 @@ public void tearDown() {
...
-    Schema schema =
-        Schema.of(
-            Field.of("stringField", LegacySQLTypeName.STRING),
-            Field.of("booleanField", LegacySQLTypeName.BOOLEAN));
-    Table table = createTableHelper(generatedDatasetName, tableName, schema);
-
     // Extract table content to GCS in CSV format
-    ExtractTableToJSON.extractTableToJSON(table, "CSV", "gs://my-bucket/extractTest.csv");
+    ExtractTableToJSON.extractTableToJSON(
+        "bigquery-public-data", "samples", "shakespeare", "gs://my-bucket/extractTest.csv");
```
kurtisvg: Is `my-bucket` a valid bucket name for these tests? Should these values be passed in via environment variables instead? (if yes, please add a check that the environment variables are set or throw an error if not)

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,26 +1,61 @@
...
     // TODO(developer): Replace these variables before running the sample.
-    Table table = null;
-    String format = "CSV";
+    String projectId = "bigquery-public-data";
+    String datasetName = "samples";
+    String tableName = "shakespeare";
     String bucketName = "my-bucket";
-    String gcsFileName = "gs://" + bucketName + "/extractTest.csv";
-    extractTableToJSON(table, format, gcsFileName);
+    String destinationUri = "gs://" + bucketName + "/path/to/file";
```
kurtisvg: nit: bucketName var is only user here - should you consider inlining it?

samples/src/main/java/com/example/bigquery/ExtractTableToJSON.java
```diff
@@ -1,26 +1,61 @@
...
+
+    // Extract table
```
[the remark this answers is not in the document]
stephaniewang526: removed empty line and unhelpful comment

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
```diff
@@ -0,0 +1,74 @@
...
+  public static void extractTableToJson(
+      String projectId, String datasetName, String tableName, String destinationUri) {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests.
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+    TableId tableId = TableId.of(projectId, datasetName, tableName);
+    Table table = bigquery.getTable(tableId);
+
+    // For more information on export format available see:
```
kurtisvg: nit:
```suggestion
    // For more information on export formats available see:
```

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
```diff
@@ -0,0 +1,74 @@
...
+    // https://cloud.google.com/bigquery/docs/exporting-data#export_formats_and_compression_types
+    // For more information on Job see:
+    // https://googleapis.dev/java/google-cloud-clients/latest/index.html?com/google/cloud/bigquery/package-summary.html
+    Job job = table.extract("CSV", destinationUri);
+    try {
+      Job completedJob = job.waitFor();
```
kurtisvg: nit: preferably this would have a comment that it's a blocking operation (may not be obvious since it doesn't follow existing Java standards).

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
```diff
@@ -0,0 +1,74 @@
...
+      if (completedJob == null) {
+        // Job no longer exists
```
kurtisvg: nit: This comment is not useful - it doesn't provide any new information about the code it's commenting on. 

Also consider updating the printlln to be more specific - What does it mean if the job no longer exists? Does this mean that it was completed? Failed? Canceled or deleted else by someone else?

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
```diff
@@ -0,0 +1,74 @@
...
+        System.out.println("Job no longer exists");
+        return;
+      } else if (completedJob.getStatus().getError() != null) {
+        // Job failed, handle error
```
kurtisvg: nit: comment not useful, consider removing

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
```diff
@@ -0,0 +1,74 @@
...
+        System.out.println(
+            "BigQuery was unable to extract due to an error: \n" + job.getStatus().getError());
+        return;
+      } else {
```
kurtisvg: nit: Consider remove else to make the successful path the "default" as I mentioned in the last comment

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
kurtisvg: (same for some of these other comments).

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
```diff
@@ -0,0 +1,74 @@
...
+        // Job completed successfully
+        System.out.println("Table export successful. Check in GCS bucket for the CSV file.");
+      }
+    } catch (InterruptedException e) {
```
kurtisvg: `InterruptedException` is a low level system error that happens when a thread is interrupted while performing a job - it should probably just be bubbled up.

What happened to BigQueryException? Is it not thrown by any of the operations?

samples/src/test/java/com/example/bigquery/ExtractTableToJsonIT.java
```diff
@@ -0,0 +1,63 @@
...
+  public void testExtractTableToJson() {
+    String projectId = "bigquery-public-data";
+    String datasetName = "samples";
+    String tableName = "shakespeare";
+    String bucketName = RemoteStorageHelper.generateBucketName();
+    String destinationUri = "gs://" + bucketName + "/extractTest.csv";
+
+    // Create GCS bucket to store extracted file
+    Storage storage = RemoteStorageHelper.create().getOptions().getService();
+    storage.create(BucketInfo.of(bucketName));
```
kurtisvg: This will lead to LOTS of different GCS buckets being created. Instead, consider using an environment variable to get the value for an existing bucket. You can use a function like [this](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/dlp/src/test/java/dlp/snippets/InspectTests.java#L39) to make it clear this resource is required to run the test.

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
stephaniewang526: I tried to expand a bit more on the println - but i think it could be due to any reason as it's not specified in the api doc either: https://googleapis.dev/java/google-cloud-clients/latest/index.html?com/google/cloud/bigquery/package-summary.html

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
stephaniewang526: The doc says the below are thrown:

Throws:
BigQueryException - upon failure, check Throwable.getCause() for details
InterruptedException - if the current thread gets interrupted while waiting for the job to complete

So I will catch both -- adding BigQueryException back

samples/src/main/java/com/example/bigquery/ExtractTableToJson.java
```diff
@@ -0,0 +1,71 @@
...
+    String destinationUri = "gs://" + bucketName + "/path/to/file";
+    extractTableToJson(projectId, datasetName, tableName, destinationUri);
+  }
+
+  // Exports datasetName:tableName to destinationUri as raw CSV
+  public static void extractTableToJson(
+      String projectId, String datasetName, String tableName, String destinationUri) {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests.
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
```
kurtisvg: All of the methods that can throw `BigQueryException` need to be in a try/catch.

`BigQueryException` is unfortunately an unchecked exception, so we need to be super careful to show users to catch it (since it's not declared in the method).

Issue #72, opened 2020-01-03
Reformat plugin is misbehaving with spaces


elharo:
The autoformat plugin is formatting code like this:

```
    expect(
            bigqueryRpcMock.open(
                new com.google.api.services.bigquery.model.Job()
                    .setJobReference(JOB_INFO.getJobId().toPb())
                    .setConfiguration(LOAD_CONFIGURATION.toPb())))
        .andThrow(ex);
```

That is, it is indenting code eight spaces on continuation lines, possibly in concert with Eclipse. I'm not sure why.

stephaniewang526:
Issue raised in google-java-format: https://github.com/google/google-java-format/issues/430.

stephaniewang526:
As noted in ticket raised -- this is expected formatting behavior. https://github.com/google/google-java-format/wiki/FAQ#google-java-format-always-follows-the-rectangle-rule-right-no-exceptions

PR #76, opened 2020-01-03
feat: new sample - Tables: Copy multiple tables

stephaniewang526:
Fixes b/147141782

### DIFF
--- a/samples/src/main/java/com/example/bigquery/CopyMultipleTables.java
+++ b/samples/src/main/java/com/example/bigquery/CopyMultipleTables.java
@@ -47,8 +47,10 @@ public class CopyMultipleTables {
       // https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/bigquery/JobConfiguration.html
       CopyJobConfiguration configuration =
           CopyJobConfiguration.newBuilder(
-              destinationTable,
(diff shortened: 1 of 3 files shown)


samples/src/main/java/com/example/bigquery/CopyMultipleTables.java
```diff
@@ -0,0 +1,76 @@
...
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      TableId destinationTable = TableId.of(destinationDatasetName, destinationTableId);
+
+      // For more information on CopyJobConfiguration see:
+      // https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/bigquery/JobConfiguration.html
+      CopyJobConfiguration configuration =
+          CopyJobConfiguration.newBuilder(
+                  destinationTable,
```
elharo: note the eight space indent here. This is inconsistent with google style that specifies four spaces for a continuation line. It's the same thing I saw this morning in a different PR and filed #72 about.

I wonder if we need to update our version of something related to formatting?

stephaniewang526: Yes - I raised a ticket with the team responsible for the tool (coveo). We do have the latest version (2.9). I will keep you posted.

PR #77, opened 2020-01-04
feat: make BQ client Autocloseable

stephaniewang526:
This will allow us to use try-with-resources in our samples.

Fixes b/145229155

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
@@ -27,7 +27,8 @@ import com.google.cloud.http.HttpTransportOptions;
 import com.google.common.collect.ImmutableSet;
 import java.util.Set;
 
-public class BigQueryOptions extends ServiceOptions<BigQuery, BigQueryOptions> implements AutoCloseable{
+public class BigQueryOptions extends ServiceOptions<BigQuery, BigQueryOptions>
+    implements AutoCloseable {
 
   private static final String API_SHORT_NAME = "BigQuery";
   private static final String BIGQUERY_SCOPE = "https://www.googleapis.com/auth/bigquery";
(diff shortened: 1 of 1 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -37,6 +38,11 @@
   // set the option ThrowNotFound when you want to throw the exception when the value not found
   private boolean setThrowNotFound;
 
+  @Override
+  public void close() throws Exception {
+    System.out.println("Closed BigQuery Resource");
```
tswast: I imagine there are some resources to actually clean up. For example, in Python there are two long-running sockets (one for handling auth token updates, one for the actual API requests) that need to be explicitly closed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -27,7 +27,8 @@
 import com.google.common.collect.ImmutableSet;
 import java.util.Set;
 
-public class BigQueryOptions extends ServiceOptions<BigQuery, BigQueryOptions> {
+public class BigQueryOptions extends ServiceOptions<BigQuery, BigQueryOptions>
```
tswast: Isn't it the Client class that needs to be autocloseable?

kurtisvg: +1 - this doesn't help users much.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
kurtisvg: +1 - Also, we probably don't want to print to `stdout`. If we do want to print this anywhere, we want to use a debug logger.

PR #82, opened 2020-01-07
feat: add enforce linkage checker

athakor:
Towards [#6396](https://github.com/googleapis/google-cloud-java/issues/6396)


BenWhitehead:
This enforcer config is already present in the shared-config since at least version 0.3.0 as can be seen [here](https://github.com/googleapis/java-shared-config/blob/v0.3.0/pom.xml#L204-L226)

PR #83, opened 2020-01-07
chore: clean up samples integration test script

stephaniewang526:


.kokoro/run_samples_tests.sh
```diff
@@ -16,42 +16,19 @@
...
-gcloud -v
 echo "********** MAVEN INFO  ***********"
 mvn -v
 
 # Setup required env variables
 source ${KOKORO_GFILE_DIR}/bigquery_secrets.txt
 echo "********** Successfully Set All Environment Variables **********"
 
-# Activate service account
-gcloud auth activate-service-account \
```
kurtisvg: I believe you should still use this to make sure you are using the correct service account and project.

PR #85, opened 2020-01-07
feat: new sample - Tables: Add empty column

stephaniewang526:
Fixes b/147297640


samples/src/main/java/com/example/bigquery/AddEmptyColumn.java
```diff
@@ -0,0 +1,72 @@
...
+      // Create a new schema adding the current fields, plus the new one
+      List<Field> fieldList = new ArrayList<Field>();
+      for (Field f : fields) {
+        fieldList.add(f);
+      }
+      fieldList.add(newField);
+      Schema newSchema = Schema.of(fieldList);
+
+      // Update the table with the new schema
+      table.toBuilder().setDefinition(StandardTableDefinition.of(newSchema)).build().update();
```
kurtisvg: nit: Consider splitting this into two actions to make more clear to users what's happening:
```java
// Update the table definition to include the new schema 
updatedTable = table.toBuilder().setDefinition(StandardTableDefinition.of(newSchema)).build();
// Push updates to the BigQuery service
updatedTable.update(); 
```

samples/src/main/java/com/example/bigquery/AddEmptyColumn.java
```diff
@@ -0,0 +1,72 @@
...
+      Schema schema = table.getDefinition().getSchema();
+      FieldList fields = schema.getFields();
+
+      // Create the new field/column
+      String randomColumnName = "new_" + UUID.randomUUID().toString().replace('-', '_');
+      Field newField = Field.of(randomColumnName, LegacySQLTypeName.STRING);
+
+      // Create a new schema adding the current fields, plus the new one
+      List<Field> fieldList = new ArrayList<Field>();
+      for (Field f : fields) {
```
kurtisvg: nit: Consider using `forEach`:
```java
List<Field> fieldList = new ArrayList<Field>()
fields.forEach(fieldList::add)
```

samples/src/main/java/com/example/bigquery/AddEmptyColumn.java
```diff
@@ -0,0 +1,72 @@
...
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      Table table = bigquery.getTable(datasetName, tableId);
+      Schema schema = table.getDefinition().getSchema();
+      FieldList fields = schema.getFields();
+
+      // Create the new field/column
+      String randomColumnName = "new_" + UUID.randomUUID().toString().replace('-', '_');
```
kurtisvg: Do we expect users to want to generate a column name with a UUID? Perhaps this would be better as a provided argument, and we can uses a UUID from the tests side instead.

samples/src/main/java/com/example/bigquery/AddEmptyColumn.java
stephaniewang526: You're right! A great place to use lambda exp!

PR #89, opened 2020-01-08
feat: new sample - Tables: Relax column

stephaniewang526:
Fixes b/147354241

### DIFF
--- a/samples/src/main/java/com/example/bigquery/RelaxColumnMode.java
+++ b/samples/src/main/java/com/example/bigquery/RelaxColumnMode.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


samples/src/main/java/com/example/bigquery/RelaxColumnMode.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public static void relaxColumnMode(String datasetName, String tableId) {
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      Table table = bigquery.getTable(datasetName, tableId);
+
+      // Create new relaxed schema based on the existing table schema
```
kurtisvg: nit: This seems to imply that you are modifying an existing Schema, but really you are rebuilding a new schema from scratch here. Would it be better to show the users how to retrieve the current Schema, and then modify the necessary fields instead of rebuilding from scratch?

PR #94, opened 2020-01-09
chore: clean up IT

stephaniewang526:


samples/src/test/java/com/example/bigquery/RelaxColumnModeIT.java
```diff
@@ -68,10 +68,6 @@ public void testRelaxColumnMode() {
         "gcloud_test_table_temp_" + UUID.randomUUID().toString().replace('-', '_');
     Schema originalSchema =
         Schema.of(
-            // The only supported modification you can make to a column's mode is changing it from
-            // REQUIRED to NULLABLE
-            // Changing a column's mode from REQUIRED to NULLABLE is also called column relaxation
-            // INFO: LegacySQLTypeName will be updated to StandardSQLTypeName in release 1.103.0
```
kurtisvg: Do you want to remove this comment? Seems still relevant to the sample.

stephaniewang526:
This is only the test not the sample so I thought I should remove it.

kurtisvg:
> This is only the test not the sample so I thought I should remove it.

Oh sorry, that totally make sense.

PR #97, opened 2020-01-10
feat: sample - run a legacy SQL query

stephaniewang526:
Fixes b/147485316

### DIFF
--- a/samples/src/main/java/com/example/bigquery/RunLegacyQuery.java
+++ b/samples/src/main/java/com/example/bigquery/RunLegacyQuery.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.bigquery;
+
+// [START bigquery_query_legacy]
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryException;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.QueryJobConfiguration;
+
+public class RunLegacyQuery {
+
+  public static void runRunLegacyQuery() {
+    runLegacyQuery();
+  }
+
+  public static void runLegacyQuery() {
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      String query =
+          "SELECT corpus FROM [bigquery-public-data:samples.shakespeare] GROUP BY corpus;";
+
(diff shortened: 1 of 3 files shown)


samples/src/main/java/com/example/bigquery/RunLegacyQuery.java
```diff
@@ -0,0 +1,49 @@
...
+
+// [START bigquery_query_legacy]
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryException;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.QueryJobConfiguration;
+
+public class RunLegacyQuery {
+
+  public static void runRunLegacyQuery() {
```
kurtisvg: I think since `RunLegacyQuery` doesn't have any arguments, you can just leave this out.

samples/src/main/java/com/example/bigquery/RunLegacyQuery.java
```diff
@@ -0,0 +1,49 @@
...
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      String query =
+          "SELECT corpus FROM [bigquery-public-data:samples.shakespeare] GROUP BY corpus;";
+
+      QueryJobConfiguration queryConfig =
+          // To use legacy SQL syntax, set useLegacySql to true.
+          QueryJobConfiguration.newBuilder(query).setUseLegacySql(true).build();
```
kurtisvg: Is this actually running the query? Or is it just configuring it? Shouldn't you need to do something to send it?

How do you process data received from the query?

samples/src/main/java/com/example/bigquery/RunLegacyQuery.java
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      String query =
+          "SELECT corpus FROM [bigquery-public-data:samples.shakespeare] GROUP BY corpus;";
+
+      QueryJobConfiguration queryConfig =
+          // To use legacy SQL syntax, set useLegacySql to true.
```
kurtisvg: nit: I would probably put this comment above the `query = ` line, since that's the point at which you want to use legacy SQL syntax

samples/src/main/java/com/example/bigquery/RunLegacyQuery.java
```diff
@@ -0,0 +1,55 @@
...
+          QueryJobConfiguration.newBuilder(query).setUseLegacySql(true).build();
+
+      // Execute the query.
+      TableResult result = bigquery.query(queryConfig);
+
+      // Print the results.
+      result
+          .iterateAll()
+          .forEach(row -> row.forEach(val -> System.out.println(val.getStringValue())));
```
kurtisvg: Better, but this will just print out a bunch of random values on separate lines. Can we preserve the rows together?

```java
System.out.println("Results")
for (FieldValueList row : result.iterateAll()) {
  row.forEach(val -> System.print(val.getStringValue() + " ")
  System.out.println();
}
```

stephaniewang526: The naming was confusing - I will change that.
This now prints out each corpus name and looks like this:
sonnets
various
1kinghenryvi

To preserve the rows, the printout would look like this:
FieldValue{attribute=PRIMITIVE, value=sonnets}
FieldValue{attribute=PRIMITIVE, value=various}
FieldValue{attribute=PRIMITIVE, value=1kinghenryvi}

I think it's better if I just print out the value.

PR #98, opened 2020-01-10
feat: sample - Get dataset properties

stephaniewang526:
Fixes b/147485983

### DIFF
--- a/samples/src/main/java/com/example/bigquery/GetDatasetInfo.java
+++ b/samples/src/main/java/com/example/bigquery/GetDatasetInfo.java
@@ -16,10 +16,14 @@
 
 package com.example.bigquery;
 
+import com.google.api.gax.paging.Page;
 import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQuery.TableListOption;
 import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetId;
+import com.google.cloud.bigquery.Table;
 
 // [START bigquery_get_dataset]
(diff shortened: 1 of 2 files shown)


samples/src/main/java/com/example/bigquery/GetDatasetInfo.java
```diff
@@ -0,0 +1,47 @@
...
+    getDatasetInfo(projectId, datasetName);
+  }
+
+  public static void getDatasetInfo(String projectId, String datasetName) {
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+      DatasetId datasetId = DatasetId.of(projectId, datasetName);
+      bigquery.getDataset(datasetId);
```
kurtisvg: Does this return a dataset? If so we should hint towards the user how they can use that. For example, we shouldn't just say "Info retrieved successfully"  but interact with the object to give the user an example of how to use the Object returned.

samples/src/main/java/com/example/bigquery/GetDatasetInfo.java
```diff
@@ -0,0 +1,63 @@
...
+      DatasetId datasetId = DatasetId.of(projectId, datasetName);
+      Dataset dataset = bigquery.getDataset(datasetId);
+
+      // View dataset properties
+      String description = dataset.getDescription();
+      System.out.println(description);
+
+      // View tables in the dataset
+      // For more information on listing tables see:
+      // https://javadoc.io/static/com.google.cloud/google-cloud-bigquery/0.22.0-beta/com/google/cloud/bigquery/BigQuery.html
```
kurtisvg: nit: Doesn't seem like a useful link - what should the user be looking for?

PR #112, opened 2020-01-16
feat: sample - add col load append

stephaniewang526:
Fixes b/147690070

### DIFF
--- a/samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
+++ b/samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
@@ -37,7 +37,7 @@ import java.util.UUID;
 
 public class AddColumnLoadAppend {
 
-  public static void runAddColumnLoadAppend() {
+  public static void runAddColumnLoadAppend() throws Exception {
     // TODO(developer): Replace these variables before running the sample.
     String datasetName = "MY_DATASET_NAME";
     String tableName = "MY_TABLE_NAME";
@@ -45,7 +45,8 @@ public class AddColumnLoadAppend {
     addColumnLoadAppend(datasetName, tableName, sourceUri);
   }
 
-  public static void addColumnLoadAppend(String datasetName, String tableName, String sourceUri) {
+  public static void addColumnLoadAppend(String datasetName, String tableName, String sourceUri)
+      throws Exception {
     try {
       // Initialize client that will be used to send requests. This client only needs to be created
       // once, and can be reused for multiple requests.
@@ -91,10 +92,9 @@ public class AddColumnLoadAppend {
       } else if (completedJob.getStatus().getError() != null) {
         // You can also look at queryJob.getStatus().getExecutionErrors() for all
         // errors, not just the latest one.
-        System.out.println(
+        throw new Exception(
             "BigQuery was unable to load into the table due to an error: \n"
                 + loadJob.getStatus().getError());
-        return;
       }
 
       System.out.println("Column successfully added during load append job");
--- a/samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
+++ b/samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
@@ -17,7 +17,6 @@
 package com.example.bigquery;
 
 // [START bigquery_relax_column_load_append]
-
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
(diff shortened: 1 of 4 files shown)


samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
```diff
@@ -0,0 +1,102 @@
...
+      Job loadJob = bigquery.create(JobInfo.newBuilder(loadJobConfig).setJobId(jobId).build());
+      System.out.println(loadJob.getJobId());
+
+      // Load data from a GCS parquet file into the table
+      // Blocks until this load table job completes its execution, either failing or succeeding.
+      Job completedJob = loadJob.waitFor();
+
+      // Check for errors
+      if (completedJob == null) {
+        throw new RuntimeException("Job no longer exists");
```
kurtisvg: nit: Probably InvalidStateException would make more sense?

RuntimeException is unchecked, making it a bad choice to suggest to users.

samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
```diff
@@ -0,0 +1,102 @@
...
+      } else if (completedJob.getStatus().getError() != null) {
+        // You can also look at queryJob.getStatus().getExecutionErrors() for all
+        // errors, not just the latest one.
+        throw new RuntimeException(loadJob.getStatus().getError().toString());
```
kurtisvg: nit: It there error throwable? e.g. `throw loadJob.getStatus().getError().toString()`? Else use a checked error.

samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
stephaniewang526: InvalidStateException is not available so I will print out the error message instead

samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
stephaniewang526: yes - in fact, it helped me debug when I wasn't able to figure out what was wrong with the job.

kurtisvg: What I meant was can you throw the Error directly - using `throw loadJob.getStatus().getError()`?

Otherwise you really shouldn't throw a unchecked exceptions like this - checked exceptions are preferred. (A regular `Exception` is considered checked, while `RuntimeExceptions` are unchecked.)

stephaniewang526: Oh I see... I'm updating this to be print it out since `loadJob.getStatus().getError()` is a String instead of Throwable.

samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
kurtisvg: I think there's a miscommunication on error handling here.

If something goes wrong, we definitely want to throw an exception - but we want a checked exception vs unchecked.

Check out this article for a breakdown of [checked v unchecked](https://howtodoinjava.com/java/exception-handling/checked-vs-unchecked-exceptions-in-java/).

samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
kurtisvg: We want to throw - it should just be a checked exception.

samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
stephaniewang526: Throwing Exception so that it's checked.

samples/src/main/java/com/example/bigquery/AddColumnLoadAppend.java
```diff
@@ -0,0 +1,106 @@
...
+      Job loadJob = bigquery.create(JobInfo.newBuilder(loadJobConfig).setJobId(jobId).build());
+      System.out.println(loadJob.getJobId());
+
+      // Load data from a GCS parquet file into the table
+      // Blocks until this load table job completes its execution, either failing or succeeding.
+      Job completedJob = loadJob.waitFor();
+
+      // Check for errors
+      if (completedJob == null) {
+        System.out.println("Job not executed since it no longer exists.");
```
kurtisvg: nit: This should probably throw and error as well.

PR #115, opened 2020-01-17
feat: add new sample - Tables: Relax column query append

stephaniewang526:
Fixes b/147808961

### DIFF
--- a/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
+++ b/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
@@ -33,7 +33,7 @@ import java.util.UUID;
 
 public class RelaxTableQuery {
 
-  public static void runRelaxTableQuery() {
+  public static void runRelaxTableQuery() throws Exception {
     // TODO(developer): Replace these variables before running the sample.
     String datasetName = "MY_DATASET_NAME";
     String tableName = "MY_TABLE_NAME";
@@ -46,7 +46,8 @@ public class RelaxTableQuery {
     relaxTableQuery(datasetName, tableName, query);
   }
 
-  public static void relaxTableQuery(String datasetName, String tableName, String query) {
+  public static void relaxTableQuery(String datasetName, String tableName, String query)
+      throws Exception {
     try {
       // Initialize client that will be used to send requests. This client only needs to be created
       // once, and can be reused for multiple requests.
@@ -72,11 +73,11 @@ public class RelaxTableQuery {
 
       // Check for errors
       if (queryJob == null) {
-        throw new RuntimeException("Job no longer exists");
+        throw new Exception("Job no longer exists");
       } else if (queryJob.getStatus().getError() != null) {
         // You can also look at queryJob.getStatus().getExecutionErrors() for all
         // errors, not just the latest one.
-        throw new RuntimeException(queryJob.getStatus().getError().toString());
+        throw new Exception(queryJob.getStatus().getError().toString());
       }
 
       // Get the results.
--- a/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
+++ b/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
@@ -35,18 +35,13 @@ public class RelaxTableQuery {
 
   public static void runRelaxTableQuery() throws Exception {
     // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
     String datasetName = "MY_DATASET_NAME";
     String tableName = "MY_TABLE_NAME";
-    String query =
-        "SELECT "
-            + "word "
-            // TODO(developer): Replace projectId.dataset.tableName with yours
-            + "FROM `java-docs-samples-testing.bigquery_test_dataset.RELAX_TABLE_QUERY_TEST` "
-            + "WHERE word like '%is%' ";
-    relaxTableQuery(datasetName, tableName, query);
+    relaxTableQuery(projectId, datasetName, tableName);
   }
 
-  public static void relaxTableQuery(String datasetName, String tableName, String query)
+  public static void relaxTableQuery(String projectId, String datasetName, String tableName)
       throws Exception {
     try {
       // Initialize client that will be used to send requests. This client only needs to be created
@@ -54,6 +49,10 @@ public class RelaxTableQuery {
       BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
 
       TableId tableId = TableId.of(datasetName, tableName);
+
+      String sourceTable = "`" + projectId + "." + datasetName + "." + tableName + "`";
+      String query =
+          "SELECT " + "word\n" + "FROM " + sourceTable + "\n" + "WHERE word like '%is%' ";
 
       QueryJobConfiguration queryConfig =
           QueryJobConfiguration.newBuilder(query)
--- a/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
+++ b/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
@@ -37,22 +37,22 @@ public class RelaxTableQuery {
     // TODO(developer): Replace these variables before running the sample.
     String datasetName = "MY_DATASET_NAME";
     String tableName = "MY_TABLE_NAME";
-    relaxTableQuery(datasetName, tableName);
+    String query =
+        "SELECT "
+            + "word "
+            // TODO(developer): Replace projectId.dataset.tableName with yours
+            + "FROM `java-docs-samples-testing.bigquery_test_dataset.RELAX_TABLE_QUERY_TEST` "
+            + "WHERE word like '%is%' ";
+    relaxTableQuery(datasetName, tableName, query);
   }
 
-  public static void relaxTableQuery(String datasetName, String tableName) {
+  public static void relaxTableQuery(String datasetName, String tableName, String query) {
     try {
       // Initialize client that will be used to send requests. This client only needs to be created
       // once, and can be reused for multiple requests.
       BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
 
       TableId tableId = TableId.of(datasetName, tableName);
-      String query =
-          "SELECT "
-              + "word "
-              // TODO(developer): Replace projectId.dataset.tableName with yours
-              + "FROM `java-docs-samples-testing.bigquery_test_dataset.RELAX_TABLE_QUERY_TEST` "
-              + "WHERE word like '%is%' ";
 
       QueryJobConfiguration queryConfig =
           QueryJobConfiguration.newBuilder(query)
--- a/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
+++ b/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
@@ -41,6 +41,7 @@ public class RelaxTableQuery {
     relaxTableQuery(projectId, datasetName, tableName);
   }
 
+  // To relax all columns in a destination table when you append data to it during a query job
   public static void relaxTableQuery(String projectId, String datasetName, String tableName)
       throws Exception {
     try {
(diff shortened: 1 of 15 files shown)


samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      TableId tableId = TableId.of(datasetName, tableName);
+      String query =
+          "SELECT "
+              + "word "
+              // TODO(developer): Replace projectId.dataset.tableName with yours
```
kurtisvg: This should be an arg that you pass in (or more preferably formatted from existing arguments)- all the //TODO's should be in the top function so they are easy to find.

Can you do something like: `String tableRef = String.format("%s.%s.%s", projectId, datasetId, tableName)`?

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+import com.google.cloud.bigquery.TableResult;
+import com.google.common.collect.ImmutableList;
+import java.util.UUID;
+
+public class RelaxTableQuery {
+
+  public static void runRelaxTableQuery() throws Exception {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String datasetName = "MY_DATASET_NAME";
```
nnegrey: @kurtisvg, do we have a convention for input fields like:
"MY_PROJECT_ID" is all caps because it is some field they have to already have

Where
"my_dataset_name" is lower case to show this is a user provided field?

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+    String tableName = "MY_TABLE_NAME";
+    relaxTableQuery(projectId, datasetName, tableName);
+  }
+
+  public static void relaxTableQuery(String projectId, String datasetName, String tableName)
+      throws Exception {
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
```
nnegrey: nit: might be worth adding the rest:

```
// Initialize client that will be used to send requests. This client only needs to be created
// once, and can be reused for multiple requests. After completing all of your requests, call
// the "close" method on the client to safely clean up any remaining background resources.
```

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String datasetName = "MY_DATASET_NAME";
+    String tableName = "MY_TABLE_NAME";
+    relaxTableQuery(projectId, datasetName, tableName);
+  }
+
+  public static void relaxTableQuery(String projectId, String datasetName, String tableName)
+      throws Exception {
+    try {
```
nnegrey: Do BigQuery samples usually not use the java `try-with-resources?

Like:
```
try (BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService()) {
...
```
https://docs.oracle.com/javase/7/docs/technotes/guides/language/try-with-resources.html

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      TableId tableId = TableId.of(datasetName, tableName);
+
+      String sourceTable = "`" + projectId + "." + datasetName + "." + tableName + "`";
+      String query =
+          "SELECT " + "word\n" + "FROM " + sourceTable + "\n" + "WHERE word like '%is%' ";
```
nnegrey: Would you want to use a StringBuilder here?

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+              .setDestinationTable(tableId)
+              .build();
+
+      // Create a job ID so that we can safely retry.
+      JobId jobId = JobId.of(UUID.randomUUID().toString());
+      Job queryJob = bigquery.create(JobInfo.newBuilder(queryConfig).setJobId(jobId).build());
+
+      queryJob = queryJob.waitFor();
+
+      // Check for errors
```
nnegrey: Do the error for BigQuery follow the same as this page?
https://cloud.google.com/apis/design/errors

If so, might be worth linking to that page unless the errors themselves are descriptive.

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+      // Get the results.
+      TableResult results = queryJob.getQueryResults();
+
+      // Print all pages of the results.
+      results
+          .iterateAll()
+          .forEach(
+              rows -> {
+                rows.forEach(row -> System.out.println("row: " + row.toString()));
+              });
```
nnegrey: @kurtisvg, is this fine or is there a preference for traditional for loops?

Like:
```
for (Row row in results.iterateAll()) { ...
```

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,99 @@
...
+
+  public static void runRelaxTableQuery() throws Exception {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String datasetName = "MY_DATASET_NAME";
+    String tableName = "MY_TABLE_NAME";
+    relaxTableQuery(projectId, datasetName, tableName);
+  }
+
+  public static void relaxTableQuery(String projectId, String datasetName, String tableName)
```
nnegrey: nit: might be worth putting a comment on what it means to relax a table query.

samples/src/test/java/com/example/bigquery/RelaxTableQueryIT.java
```diff
@@ -0,0 +1,88 @@
...
+                .build());
+
+    CreateTable.createTable(BIGQUERY_DATASET_NAME, tableName, originalSchema);
+
+    RelaxTableQuery.relaxTableQuery(BIGQUERY_PROJECT_ID, BIGQUERY_DATASET_NAME, tableName);
+    assertThat(bout.toString())
+        .contains("Successfully relaxed all columns in destination table during query job");
+
+    // Clean up
+    DeleteTable.deleteTable(BIGQUERY_DATASET_NAME, tableName);
```
nnegrey: nit: should this be in `tearDown`?

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
stephaniewang526: I think they're both fields to be provided by user - MY_PROJECT_ID is already set only for testing. If they just run the sample, they will still need to provide the projectId.

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
stephaniewang526: Unfortunately, BigQuery is Apiary generated and does not have a close method I/the user could call.

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
kurtisvg: forEach is preffered over traditional for-loops because it allows for lambdas (it's backwards in the sample guide atm)

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
stephaniewang526: The errors for Jobs don't have error codes. I think they're decently descriptive.

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
kurtisvg: Java will actually [optimize concats like this](https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java) - you only need to use StringBuilder if you are appending iteratively

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
kurtisvg: We don't have a convention AFIAK - I prefer `"ALL-CAPS-WITH-HYPENS"` but it's not really a significant difference

samples/src/test/java/com/example/bigquery/RelaxTableQueryIT.java
stephaniewang526: I don't seem to be able to pass in `tableName` in tearDown() @kurtisvg thoughts?

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
stephaniewang526: That's right -- BQ client is not Autocloseable and cannot be changed into one since there's no close() method we can call to clean up resource. The Apiary client generator would have to change... something I discussed with Les and will put on hold for now.

samples/src/test/java/com/example/bigquery/RelaxTableQueryIT.java
kurtisvg: It's probably a good practice because it will ensure the tearDown occurs if the test fails. You should also use a UUID, so that multiple test runs don't step on each other. 

You can accomplish this be having `tableName` be a class field, and using `setUp` to create it and `tearDown` to destroy it.

kurtisvg: It's probably a good practice because it will ensure the tearDown occurs if the test fails. You should also use a UUID, so that multiple test runs don't step on each other. 

You can accomplish this be having `tableName` be a class field, and using `setUp` to create it and `tearDown` to destroy it.

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
kurtisvg: The BQ client is hand written and has a few quirks - one is the client isn't autoclosable.

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
kurtisvg: Another BQ client quirk - it doesn't return errors in the best way. This was a compromised we reached.

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
stephaniewang526: will do

samples/src/test/java/com/example/bigquery/RelaxTableQueryIT.java
stephaniewang526: Will do

samples/src/test/java/com/example/bigquery/RelaxTableQueryIT.java
```diff
@@ -0,0 +1,88 @@
...
+    requireEnvVar("BIGQUERY_PROJECT_ID");
+    requireEnvVar("BIGQUERY_DATASET_NAME");
+  }
+
+  @Before
+  public void setUp() throws Exception {
+    bout = new ByteArrayOutputStream();
+    out = new PrintStream(bout);
+    tableName =
+        "RELAX_TABLE_QUERY_TEST" + UUID.randomUUID().toString().substring(0, 5).replace('-', '_');
```
nnegrey: I'm guessing table name has a limit, to get more randomization would it be better to do:
```
tableName = "TEST_" + UUID.randomUUID().toString().replace('-', '').substring(0, 23);
```

PR #118, opened 2020-01-21
feat: new sample - Table: Column Based Time Partitioning

stephaniewang526:
Fixes b/147905712

### DIFF
--- a/samples/src/main/java/com/example/bigquery/CreatePartitionedTable.java
+++ b/samples/src/main/java/com/example/bigquery/CreatePartitionedTable.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


samples/src/main/java/com/example/bigquery/CreatePartitionedTable.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.cloud.bigquery.TableInfo;
+import com.google.cloud.bigquery.TimePartitioning;
+
+public class CreatePartitionedTable {
+
+  public static void runCreatePartitionedTable() {
+    // TODO(developer): Replace these variables before running the sample.
+    String datasetName = "MY_DATASET_NAME";
+    String tableName = "MY_TABLE_NAME";
+    Schema schema =
```
kurtisvg: The schema should probably be defined as part of the main sample - it's not required for testing.

PR #119, opened 2020-01-22
feat: add new sample - Load: Column Based Time Partitioning

stephaniewang526:
Fixes b/148144641

### DIFF
--- a/samples/src/main/java/com/example/bigquery/LoadPartitionedTable.java
+++ b/samples/src/main/java/com/example/bigquery/LoadPartitionedTable.java
@@ -16,7 +16,7 @@
 
 package com.example.bigquery;
 
-// [START bigquery_relax_column_load_append]
+// [START bigquery_load_table_partitioned]
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.BigQueryOptions;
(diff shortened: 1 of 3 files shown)


samples/src/main/java/com/example/bigquery/LoadPartitionedTable.java
```diff
@@ -0,0 +1,100 @@
...
+              Field.of("name", StandardSQLTypeName.STRING),
+              Field.of("post_abbr", StandardSQLTypeName.STRING),
+              Field.of("date", StandardSQLTypeName.DATE));
+
+      // Configure time partitioning. For full list of options, see:
+      // https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TimePartitioning
+      TimePartitioning partitioning =
+          TimePartitioning.newBuilder(TimePartitioning.Type.DAY)
+              .setField("date")
+              .setExpirationMs(7776000000L)
```
kurtisvg: What is the significance of this value? Can you use `java.time` instead?

```java
.setExpirationMs(Duration.of(1, ChronoUnit.DAY).toMillis()); 
```

stephaniewang526: using java.time here instead.

samples/src/test/java/com/example/bigquery/LoadPartitionedTableIT.java
```diff
@@ -0,0 +1,72 @@
...
+
+    String tableName = "PARTITIONED_TABLE_TEST";
+
+    LoadPartitionedTable.loadPartitionedTable(BIGQUERY_DATASET_NAME, tableName, sourceUri);
+
+    assertThat(bout.toString())
+        .contains("Data successfully loaded into time partitioned table during load job");
+
+    // Clean up
+    DeleteTable.deleteTable(BIGQUERY_DATASET_NAME, tableName);
```
kurtisvg: nit: This would be another example of where it might be better to create/delete the table as setUp and tearDown functions.

PR #121, opened 2020-01-23
feat: new sample - Query: Clustered Table

stephaniewang526:
Fixes b/148165984


samples/src/main/java/com/example/bigquery/QueryClusteredTable.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.cloud.bigquery.TableResult;
+
+public class QueryClusteredTable {
+
+  public static void runQueryClusteredTable() throws Exception {
+    // TODO(developer): Replace these variables before running the sample.
+    String query =
+        "SELECT word, word_count\n"
+            // TODO: Replace below with your partitioned and clustered table:
+            // projectId.datasetName.tableName
```
kurtisvg: This should be a provided argument - all TODO's should be in the no-args function.

PR #124, opened 2020-01-23
feat: new sample - Get Model

stephaniewang526:
Fixes b/148225513

### DIFF
--- a/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
+++ b/samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
@@ -33,7 +33,7 @@ import java.util.UUID;
 
 public class RelaxTableQuery {
 
-  public static void runRelaxTableQuery() {
+  public static void runRelaxTableQuery() throws Exception {
     // TODO(developer): Replace these variables before running the sample.
     String datasetName = "MY_DATASET_NAME";
     String tableName = "MY_TABLE_NAME";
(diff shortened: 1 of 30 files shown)


samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,100 @@
...
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      TableId tableId = TableId.of(datasetName, tableName);
+
+      String sourceTable = "`" + projectId + "." + datasetName + "." + tableName + "`";
+      String query =
+          "SELECT " + "word\n" + "FROM " + sourceTable + "\n" + "WHERE word like '%is%' ";
```
kurtisvg:
```suggestion
          "SELECT word FROM " + sourceTable + " WHERE word like '%is%' ";
```

samples/src/main/java/com/example/bigquery/RelaxTableQuery.java
```diff
@@ -0,0 +1,100 @@
...
+              // Use standard SQL syntax for queries.
+              // See: https://cloud.google.com/bigquery/sql-reference/
+              .setUseLegacySql(false)
+              .setSchemaUpdateOptions(ImmutableList.of(SchemaUpdateOption.ALLOW_FIELD_RELAXATION))
+              .setWriteDisposition(WriteDisposition.WRITE_APPEND)
+              .setDestinationTable(tableId)
+              .build();
+
+      // Create a job ID so that we can safely retry.
+      JobId jobId = JobId.of(UUID.randomUUID().toString());
```
kurtisvg: This looks out of place - What does this mean?  Where is the retry happening? Is the jobId required - if so, would it be better to set a specific one?

PR #125, opened 2020-01-24
feat: new sample - Models: List model resources contained in a dataset

stephaniewang526:
Fixes b/148282094

### DIFF
--- a/samples/src/main/java/com/example/bigquery/ListModels.java
+++ b/samples/src/main/java/com/example/bigquery/ListModels.java
@@ -43,7 +43,9 @@ public class ListModels {
         System.out.println("Dataset does not contain any models.");
         return;
       }
-      models.iterateAll().forEach(model -> System.out.printf("Success! Model ID: %s", model.getModelId()));
+      models
+          .iterateAll()
+          .forEach(model -> System.out.printf("Success! Model ID: %s", model.getModelId()));
     } catch (BigQueryException e) {
       System.out.println("Models not listed in dataset due to error: \n" + e.toString());
     }
--- a/samples/src/main/java/com/example/bigquery/ListModels.java
+++ b/samples/src/main/java/com/example/bigquery/ListModels.java
@@ -38,16 +38,16 @@ public class ListModels {
       // once, and can be reused for multiple requests.
       BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
 
-      try {
-        Page<Model> models = bigquery.listModels(datasetName, ModelListOption.pageSize(100));
+      Page<Model> models = bigquery.listModels(datasetName, ModelListOption.pageSize(100));
+      if (models == null) {
+        System.out.println("Dataset does not contain any models");
+      } else {
         models
             .iterateAll()
             .forEach(
                 model ->
                     System.out.println(
                         model.getModelId() + " models in dataset listed successfully."));
-      } catch (NullPointerException e) {
-        throw new NullPointerException("Dataset does not contain any models");
       }
     } catch (BigQueryException e) {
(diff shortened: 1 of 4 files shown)


samples/src/main/java/com/example/bigquery/ListModels.java
```diff
@@ -0,0 +1,57 @@
...
+
+      try {
+        Page<Model> models = bigquery.listModels(datasetName, ModelListOption.pageSize(100));
+        models
+            .iterateAll()
+            .forEach(
+                model ->
+                    System.out.println(
+                        model.getModelId() + " models in dataset listed successfully."));
+      } catch (NullPointerException e) {
```
kurtisvg: Is this if models is null? We should check models if it's`null` and (and throw an error if it is) instead of try/catching

stephaniewang526: okay

samples/src/main/java/com/example/bigquery/ListModels.java
```diff
@@ -0,0 +1,57 @@
...
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      try {
+        Page<Model> models = bigquery.listModels(datasetName, ModelListOption.pageSize(100));
+        models
+            .iterateAll()
+            .forEach(
+                model ->
+                    System.out.println(
+                        model.getModelId() + " models in dataset listed successfully."));
```
kurtisvg: This doesn't look right - this is printing as `"$ModelId models in dataset listed successfully"`.

I think you want probably want to do is print something like `Model: modelId = $modelId, modelName = $modelName

samples/src/main/java/com/example/bigquery/ListModels.java
kurtisvg: You shouldn't catch NullPointers - prevent it from happening instead.

samples/src/main/java/com/example/bigquery/ListModels.java
stephaniewang526: Looks like there's only getModelId method to get the model Id. I think I will leave it this way but update to just "model" (instead of "models").

samples/src/main/java/com/example/bigquery/ListModels.java
stephaniewang526: ok

samples/src/main/java/com/example/bigquery/ListModels.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public static void listModels(String datasetName) {
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      Page<Model> models = bigquery.listModels(datasetName, ModelListOption.pageSize(100));
+      if (models == null) {
+        System.out.println("Dataset does not contain any models");
```
kurtisvg: nit: You should use if/break over if/else here. It simplifies the control flow and makes it easier to read, and easier for people to write code after the if/else block that might rely on `models` (since it's guaranteed to be non-null at that point). 

```java
if (models == null) {
  System.out.println("Dataset does not contain any models.");
  return;
}
System.out.printf("Dataset contains %s models:\n", models.getLength());
models
  .iterateAll()
  .forEach(
     model -> System.out.printf("Model ID: %s", model.getModelId()));
```

stephaniewang526: okay - updated

PR #128, opened 2020-01-27
chore: add debug statement to StandardTableDefinition

stephaniewang526:
Debugs b/148288889

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
@@ -289,6 +289,7 @@ public abstract class StandardTableDefinition extends TableDefinition {
                       + " in table "
                       + tablePb.getTableReference())
               .initCause(e);
+          //catching Throwable due to adding initCause
         } catch (Throwable throwable) {
           throwable.printStackTrace();
         }
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
@@ -276,23 +276,26 @@ public abstract class StandardTableDefinition extends TableDefinition {
         builder.setTimePartitioning(TimePartitioning.fromPb(tablePb.getTimePartitioning()));
       } catch (IllegalArgumentException e) {
         throw new IllegalArgumentException(
-            "Got unexpected time partitioning "
-                + tablePb.getTimePartitioning()
+            "Illegal Argument - Got unexpected time partitioning "
+                + tablePb.getTimePartitioning().getType()
+                + " in project "
+                + tablePb.getTableReference().getProjectId()
+                + " in dataset "
+                + tablePb.getTableReference().getDatasetId()
                 + " in table "
-                + tablePb.getTableReference(),
+                + tablePb.getTableReference().getTableId(),
             e);
       } catch (NullPointerException e) {
-        try {
-          throw new NullPointerException(
-                  "Got unexpected time partitioning "
-                      + tablePb.getTimePartitioning()
-                      + " in table "
-                      + tablePb.getTableReference())
-              .initCause(e);
-          // catching Throwable due to adding initCause
-        } catch (Throwable throwable) {
-          throwable.printStackTrace();
-        }
+        throw new NullPointerException(
+            "Null pointer - Got unexpected time partitioning "
+                + tablePb.getTimePartitioning().getType()
+                + " in project "
+                + tablePb.getTableReference().getProjectId()
+                + " in dataset "
+                + tablePb.getTableReference().getDatasetId()
+                + " in table "
+                + tablePb.getTableReference().getTableId()
+                + e.toString());
       }
     }
     if (tablePb.getRangePartitioning() != null) {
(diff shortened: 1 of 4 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableDefinition.java
```diff
@@ -145,7 +146,7 @@ Table toPb() {
   }
 
   @SuppressWarnings("unchecked")
-  static <T extends TableDefinition> T fromPb(Table tablePb) {
+  static <T extends TableDefinition> T fromPb(Table tablePb) throws IOException, Throwable {
```
tswast: Ditto. I'm concerned that this might be a breaking change. I guess it's okay since it's protected?

Let's be more specific than "Throwable" and use IllegalArgumentException, NullPointerException.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
```diff
@@ -272,7 +272,27 @@ static StandardTableDefinition fromPb(Table tablePb) {
...
+      } catch (NullPointerException e) {
+        try {
+          throw new NullPointerException(
+                  "Got unexpected time partitioning "
+                      + tablePb.getTimePartitioning()
+                      + " in table "
+                      + tablePb.getTableReference())
+              .initCause(e);
+        } catch (Throwable throwable) {
+          throwable.printStackTrace();
```
tswast: Are we expecting to get errors when constructing the exception?

tswast: Or is this due to the `initCause`? If so, let's add a comment.

stephaniewang526: due to initCause - added comment

stephaniewang526: No - this is for .initiCause

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/StandardTableDefinitionTest.java
```diff
@@ -118,6 +125,12 @@ public void testToAndFromPb() {
         definition, TableDefinition.<StandardTableDefinition>fromPb(definition.toPb()));
   }
 
+  @Test(expected = IllegalArgumentException.class)
+  public void testInvalidTableDef() {
+    // not supposed to have this time partitioning type
+    StandardTableDefinition.fromPb(INVALID_TABLE_DEFINITION.toPb());
+  }
+
```
pmakani: use try/catch and assert.fail()

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
```diff
@@ -272,7 +272,31 @@ static StandardTableDefinition fromPb(Table tablePb) {
       builder.setStreamingBuffer(StreamingBuffer.fromPb(tablePb.getStreamingBuffer()));
     }
     if (tablePb.getTimePartitioning() != null) {
-      builder.setTimePartitioning(TimePartitioning.fromPb(tablePb.getTimePartitioning()));
+      try {
+        builder.setTimePartitioning(TimePartitioning.fromPb(tablePb.getTimePartitioning()));
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException(
+            "Illegal Argument - Got unexpected time partitioning "
+                + tablePb.getTimePartitioning().getType()
```
tswast: Actually, can we output the whole `timePartitioning` pretty-printed to a string?

tswast: That'll give us a little more context

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
```diff
@@ -272,7 +272,31 @@ static StandardTableDefinition fromPb(Table tablePb) {
...
+                + tablePb.getTableReference().getProjectId()
+                + " in dataset "
+                + tablePb.getTableReference().getDatasetId()
+                + " in table "
+                + tablePb.getTableReference().getTableId(),
+            e);
+      } catch (NullPointerException e) {
+        throw new NullPointerException(
+            "Null pointer - Got unexpected time partitioning "
+                + tablePb.getTimePartitioning().getType()
```
tswast: Same here. Let's call getTimePartitioning().toPrettyString()

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
stephaniewang526: changing to using toString()

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
stephaniewang526: changing to using toString() since toPrettyString calls for adding IOException to method signature

PR #158, opened 2020-02-06
feat: expose requirepartitionfilter field in table

### LINKED ISSUE #156: Bigquery: expose requirePartitionFilter field in Table.
- As of now requirePartitionFilter field available in TimePartitioning but there is an another 
  type of partition available which is RangePartitioning and it's already implemented. 
  so we will have to expose requirePartitionFilter at top level in the table. 
- The existing requirePartitionFilter field will be deprecated in TimePartitioning.

pmakani:
Fixes #156


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
```diff
@@ -130,6 +131,8 @@ public Table apply(TableInfo tableInfo) {
     public abstract TableInfo build();
 
     public abstract Builder setEncryptionConfiguration(EncryptionConfiguration configuration);
+
+    public abstract Builder setRequirePartitionFilter(Boolean requirePartitionFilter);
```
elharo: This is an API breaking change that requires a major version bump per semver.  If this were non-abstract with a default implementation that would not be so.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Table.java
```diff
@@ -139,7 +139,7 @@ public Builder setDefinition(TableDefinition definition) {
     }
 
     @Override
-    public TableInfo.Builder setEncryptionConfiguration(EncryptionConfiguration configuration) {
+    public Builder setEncryptionConfiguration(EncryptionConfiguration configuration) {
```
elharo: did the return type change here or did you just remove the outer class?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
pmakani: @elharo please take a look on this existing PR Ref https://github.com/googleapis/google-cloud-java/pull/6321/files#diff-2814f0d886d621f57ac6eaaabd329765R127-R134, same thing introduced before and it did not break any API change.

elharo: This breaks all existing subclasses. It looks like an incompatible API change. So does the one you cite. What am I missing?

pmakani: @shollyman do you think if users are extending TableInfo.Builder ? i think it's most unlikely that's why i introduced setter.

pmakani: @elharo TableInfo.Builder is extended only once within the client.

elharo: The class is non-final, public, and has an implicit public constructor. In hindsight, I suspect that was a mistake but we're stuck with it now until the next major version bump. We've made a commitment to our users not to break the API until then so we need to find another way around this. There are several. The cleanest and simplest is probably a default implementation that makes the method non-abstract.

shollyman: If I'm parsing elharo's comment correctly, the issue is we're starting with a new abstract implementation that all implementors must deal with, and that may break existing subclassing.

The default implementation (in the backend) treats this as false when not specified.  However, an alternate default implementation choice here may be to surface a NotImplementedException to signal to third parties about missing functionality.  Which feels like the better solution here to you @elharo?

elharo: Can the default implementation simply set the field in the usual way?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Table.java
stephaniewang526: @pmakani hi Praful - could you explain why TableInfo need to be removed here please?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
stephaniewang526: Discussed w Seth - from BQ product perspective we've been adhering to using abstract class without default, hence we will make it different here. Although point brought up here is completely valid.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Table.java
stephaniewang526: caught up w Praful -- this is to make the method consistent with the other existing methods which are not using TableInfo as an outer class.

elharo: Looking at this again. This is not OK absent a major version bump, and it probably doesn't justify that. I would mark this method deprecated and introduce a new method that returns a Table.Builder, but you don't need to delete this one now.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
stephaniewang526: will add an empty default implementation.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
```diff
@@ -399,6 +414,14 @@ public BigInteger getNumRows() {
     return labels.userMap();
   }
 
+  /**
+   * Returns true if a partition filter (that can be used for partition elimination) is required to
```
elharo: delete "to be specified"

stephaniewang526: will do once GH is back up

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
```diff
@@ -399,6 +414,14 @@ public BigInteger getNumRows() {
     return labels.userMap();
   }
 
+  /**
+   * Returns true if a partition filter (that can be used for partition elimination) is
```
stephaniewang526: @pmakani - hey i can't seem to get this right on GH... could you please see to getting the comment here right to be "is required for..."

PR #197, opened 2020-02-27
chore: move autovalue version properties from the JDK 1.8 profile into the base profile

netdpb:
This might help when users' build tools don't understand the Maven profiles configuration.


pmakani:
@googlebot I consent.

PR #198, opened 2020-02-28
chore: remove unused samples.cfg

stephaniewang526:
we have java8/java11-samples.cfg that are being used instead


stephaniewang526:
Closing due to the fact that they're auto-generated

PR #214, opened 2020-03-06
samples: new sample bigquery_auth_drive_scope

stephaniewang526:
Fixes b/146925135

### DIFF
--- a/samples/src/main/java/com/example/bigquery/AuthDriveScope.java
+++ b/samples/src/main/java/com/example/bigquery/AuthDriveScope.java
@@ -30,11 +30,12 @@ public class AuthDriveScope {
   public static void setAuthDriveScope() throws IOException {
     // Create credentials with Drive & BigQuery API scopes.
     // Both APIs must be enabled for your project before running this code.
-    String BIGQUERY_SCOPE = "https://www.googleapis.com/auth/bigquery";
-    String GOOGLE_DRIVE_SCOPE = "https://www.googleapis.com/auth/drive";
     GoogleCredentials credentials =
         ServiceAccountCredentials.getApplicationDefault()
-            .createScoped(ImmutableSet.of(BIGQUERY_SCOPE, GOOGLE_DRIVE_SCOPE));
+            .createScoped(
+                ImmutableSet.of(
+                    "https://www.googleapis.com/auth/bigquery",
+                    "https://www.googleapis.com/auth/drive"));
 
     // Instantiate a client.
     BigQuery bigquery =
--- a/samples/src/main/java/com/example/bigquery/AuthDriveScope.java
+++ b/samples/src/main/java/com/example/bigquery/AuthDriveScope.java
@@ -45,7 +45,7 @@ public class AuthDriveScope {
     System.out.println("Auth succeeded with multiple scopes. Datasets:");
     for (Dataset dataset : bigquery.listDatasets().iterateAll()) {
       System.out.printf(
-          "Auth succeeded with multiple scopes. Dataset %s%n", dataset.getDatasetId().getDataset());
+          "Dataset: %s%n", dataset.getDatasetId().getDataset());
     }
   }
 }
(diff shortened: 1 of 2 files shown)


samples/src/main/java/com/example/bigquery/AuthDriveScope.java
```diff
@@ -0,0 +1,52 @@
...
+                ImmutableSet.of(
+                    "https://www.googleapis.com/auth/bigquery",
+                    "https://www.googleapis.com/auth/drive"));
+
+    // Instantiate a client.
+    BigQuery bigquery =
+        BigQueryOptions.newBuilder().setCredentials(credentials).build().getService();
+
+    // Use the client.
+    System.out.println("Datasets:");
```
kurtisvg: Doesn't really make sense to print this multiple times. 
```suggestion
    System.out.println("Auth succeeded with multiple scopes. Datasets:");
```

samples/src/main/java/com/example/bigquery/AuthDriveScope.java
```diff
@@ -0,0 +1,52 @@
...
+    for (Dataset dataset : bigquery.listDatasets().iterateAll()) {
+      System.out.printf(
+          "Auth succeeded with multiple scopes. Dataset %s%n", dataset.getDatasetId().getDataset());
```
kurtisvg:
```suggestion
          "Dataset: %s%n", dataset.getDatasetId().getDataset());
```

samples/src/main/java/com/example/bigquery/AuthDriveScope.java
```diff
@@ -0,0 +1,52 @@
...
+    // Create credentials with Drive & BigQuery API scopes.
+    // Both APIs must be enabled for your project before running this code.
+    GoogleCredentials credentials =
+        ServiceAccountCredentials.getApplicationDefault()
+            .createScoped(
+                ImmutableSet.of(
+                    "https://www.googleapis.com/auth/bigquery",
+                    "https://www.googleapis.com/auth/drive"));
+
+    // Instantiate a client.
```
kurtisvg: Can we use the wording on the other samples? About reusing the client for multiple requests?

samples/src/main/java/com/example/bigquery/AuthDriveScope.java
```diff
@@ -0,0 +1,52 @@
...
+    BigQuery bigquery =
+        BigQueryOptions.newBuilder().setCredentials(credentials).build().getService();
+
+    // Use the client.
+    System.out.println("Datasets:");
+    for (Dataset dataset : bigquery.listDatasets().iterateAll()) {
+      System.out.printf(
+          "Auth succeeded with multiple scopes. Dataset %s%n", dataset.getDatasetId().getDataset());
+    }
+  }
```
kurtisvg: Shouldn't this catch the `BigQueryException` error?

samples/src/main/java/com/example/bigquery/AuthDriveScope.java
stephaniewang526: yes - will do

samples/src/main/java/com/example/bigquery/AuthDriveScope.java
stephaniewang526: yes - agree

samples/src/main/java/com/example/bigquery/AuthDriveScope.java
```diff
@@ -0,0 +1,57 @@
...
+          BigQueryOptions.newBuilder().setCredentials(credentials).build().getService();
+
+      // Use the client.
+      System.out.println("Auth succeeded with multiple scopes. Datasets:");
+      for (Dataset dataset : bigquery.listDatasets().iterateAll()) {
+        System.out.printf("Dataset: %s%n", dataset.getDatasetId().getDataset());
+      }
+    } catch (BigQueryException e) {
+      System.out.println("Auth failed due to error: \n" + e.toString());
+    }
```
lesv: Ok, but if you aren't handling it, and not providing any useful advice, you probably should let the higher level handle.

PR #223, opened 2020-03-16
feat: add struct query parameters

### LINKED ISSUE #203: BigQuery: Add support for struct query parameters
`QueryParameterValue` does not currently support STRUCT parameters.

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/0f6a47e19cc51a8baa9e13d9f1bb8d23c105e00a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java#L58

See the API docs: https://cloud.google.com/bigquery/querying-data#using_structs_in_parameterized_queries for how these work in the API and an example query.

pmakani:

Fixes #203

### DIFF
--- a/google-cloud-bigquery/clirr-ignored-differences.xml
+++ b/google-cloud-bigquery/clirr-ignored-differences.xml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- see http://mojo.codehaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
-<differences>
-  <difference>
-    <className>com/google/cloud/bigquery/QueryParameterValue</className>
-    <method>java.util.Map getStructTypes()</method>
-    <differenceType>7013</differenceType>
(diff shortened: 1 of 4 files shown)


google-cloud-bigquery/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://mojo.codehaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <className>com/google/cloud/bigquery/QueryParameterValue</className>
+    <method>java.util.Map getStructTypes()</method>
+    <differenceType>7013</differenceType>
```
stephaniewang526: do we have to add an abstract method? I think it may not be worth doing a major bump for this change. can we add a regular method w default implementation returning "this"? thank you!

PR #231, opened 2020-03-18
chore: skip samples module in synth

stephaniewang526:


synth.py
```diff
@@ -18,6 +18,7 @@
 
 java.common_templates(excludes=[
     'README.md',
+    '.samples',
```
chingor13: I don't think you want the `.`

stephaniewang526: ah.. yes you're right.

synth.py
```diff
@@ -18,6 +18,7 @@
 
 java.common_templates(excludes=[
     'README.md',
+    'samples',
```
chingor13:
```suggestion
    'samples/*',
```
synthtool uses glob format, we'll need to add the *

PR #236, opened 2020-03-20
feat: add support for Hive partitioning options when creating external tables

### LINKED ISSUE #235: Need support for Hive partitioning options when creating external table
The SDK doesn't support setting Hive partitioning options when creating an external table in BigQuery. Here is the corresponding API doc: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#HivePartitioningOptions.

ekanth:

Fixes #235 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
@@ -135,6 +135,14 @@ public abstract class ExternalTableDefinition extends TableDefinition {
     /** Sets the table schema. */
     public abstract Builder setSchema(Schema schema);
 
+    /** Sets the table Hive partitioning options. */
+    public Builder setHivePartitioningOptions(HivePartitioningOptions hivePartitioningOptions) {
+      return setHivePartitioningOptionsInner(hivePartitioningOptions);
+    };
+
+    abstract Builder setHivePartitioningOptionsInner(
+        HivePartitioningOptions hivePartitioningOptions);
+
     /** Creates an {@code ExternalTableDefinition} object. */
     @Override
     public abstract ExternalTableDefinition build();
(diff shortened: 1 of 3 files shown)


ekanth:
@googlebot I signed it!

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
```diff
@@ -212,6 +215,13 @@ public Boolean ignoreUnknownValues() {
   @Nullable
   public abstract Boolean getAutodetect();
 
+  /**
+   * [Experimental] Returns the HivePartitioningOptions when the data layout follows Hive partitioning convention
+   */
+  @SuppressWarnings("unchecked")
+  @Nullable
+  public abstract HivePartitioningOptions getHivePartitioningOptions();
```
stephaniewang526: this breaks binary compatibility

stephaniewang526:
In general, when you make a commit please be sure to:

1) make initial commit message in the format - "feat: blah blah..." (note that "blah" has a space in front of the colon and is in lowercase)

2) run maven coveo code formatter so that code format is correct: `mvn com.coveo:fmt-maven-plugin:format`

3) check and make sure all builds succeed and correct any build failures. In this case, abstract classes were added which break binary compatibility -- 
```
[ERROR] 7013: com.google.cloud.bigquery.ExternalTableDefinition: Abstract method 'public com.google.cloud.bigquery.HivePartitioningOptions getHivePartitioningOptions()' has been added
[ERROR] 7013: com.google.cloud.bigquery.ExternalTableDefinition$Builder: Abstract method 'public com.google.cloud.bigquery.ExternalTableDefinition$Builder setHivePartitioningOptions(com.google.cloud.bigquery.HivePartitioningOptions)' has been added
```
we don't want to do this because this renders our client unstable and would break customer's existing implementation. I suggest to try either implement without abstract and add default implementation or encapsulate with an inner method.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
ekanth: Thanks @stephaniewang526 for the review. I will update the patch and send again.

ekanth: @stephaniewang526, @pmakani I updated the patch and I wonder if the tests would automatically run?

ekanth: @stephaniewang526, thanks for triggering the tests. It was my bad with the previous patch - I missed a line that I fixed locally related to the test failures. Could you please trigger the tests again?

ekanth: Thanks @stephaniewang526. All tests passed now.

PR #248, opened 2020-03-26
samples: add multi-module samples module to bq client

stephaniewang526:
enable testing against SNAPSHOT


.kokoro/run_samples_tests.sh
```diff
@@ -32,7 +32,7 @@ gcloud auth activate-service-account \
 # Get the directory of the build script
 scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
 # Move into the samples directory
-cd ${scriptDir}/../samples/
+cd ${scriptDir}/../samples/snippets
```
chingor13: This should probably be left alone. From the samples folder, it will run all 3 suites.

samples/snapshot/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>1.0.1</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+  <!-- {x-version-update-end} -->
```
chingor13: This needs to go immediately after the google-cloud-bigquery dependency or it will end up updating all the `<version>` tags inside when doing a release.

chingor13:
Also, this shouldn't affect the library release so let's make sure this is either a `chore:` or `samples:` type

samples/snapshot/pom.xml
stephaniewang526: done

.kokoro/run_samples_tests.sh
stephaniewang526: done

samples/pom.xml
```diff
@@ -1,23 +1,31 @@
+<?xml version='1.0' encoding='UTF-8'?>
 <!--
-  ~ Copyright 2019 Google LLC
+  ~ Copyright 2020 Google LLC
```
chingor13: this one should be reverted

stephaniewang526: done

PR #323, opened 2020-05-05
feat: switch to use java-shared-deps

stephaniewang526:


pom.xml
```diff
@@ -64,16 +64,7 @@
     <github.global.server>github</github.global.server>
     <site.installationModule>google-cloud-bigquery-parent</site.installationModule>
     <google.core.version>1.93.4</google.core.version>
```
chingor13, line 66: Looks like you can remove this one too

stephaniewang526: unforch no due to https://github.com/googleapis/java-bigquery/blob/b24d6354196724195aa810fe50c6a006f04b41c7/google-cloud-bigquery/pom.xml#L65

PR #336, opened 2020-05-11
feat: add hourly partitioning support

stephaniewang526:
Fixes b/154849483

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
@@ -36,10 +36,10 @@ public abstract class TimePartitioning implements Serializable {
 
   private static final long serialVersionUID = -8565064035346940951L;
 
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
```diff
@@ -37,23 +37,23 @@
   private static final long serialVersionUID = -8565064035346940951L;
 
   /**
-   * The type of time partitioning. Currently, the only type supported is {@code DAY}, which will
-   * generate one partition per day based on data loading time.
+   * The supported types are DAY, which will generate one partition per day, and HOUR, which will
+   * generate one partition per hour. (-- Providing an empty string used to cause an error, but in
```
shollyman, line 41: The (-- --) are internal proto comments, should not be reflected in the public client docs.

stephaniewang526: done

PR #360, opened 2020-05-18
docs(samples): Added UpdateTableDescription example.

ejdarrow:
@stephaniewang526


stephaniewang526:
in the future, commit msg should be "samples: added updatetable.." so that we can pass conventionalcommit check.

PR #393, opened 2020-05-28
docs(samples): Convert tightly coupled local variable involved options inside of method into method arguments

### LINKED ISSUE #394: docs(samples): Convert tightly coupled local variable involved options inside of method into method arguments
**Is your feature request related to a problem? Please describe.**
As of the current version we can see that more options related to argument
are written inside of the test instead of included as arguments

**Describe the solution you'd like**
As this is tightly coupled with argument, it will be better to pass it as well.

**Describe alternatives you've considered**
N/A

**Additional context**
Any other context or screenshots about the feature request.

irvifa:
As of the current version we can see that more options related to argument
are written inside of the test instead of included as arguments, so I think
as this is tightly coupled with argument, it will be better to pass it as well.

Fixes #394 ☕️

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/CreatePartitionedTable.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/CreatePartitionedTable.java
@@ -34,10 +34,15 @@ public class CreatePartitionedTable {
     // TODO(developer): Replace these variables before running the sample.
     String datasetName = "MY_DATASET_NAME";
     String tableName = "MY_TABLE_NAME";
-    createPartitionedTable(datasetName, tableName);
+    Schema schema =
+            Schema.of(
+                    Field.of("stringField", StandardSQLTypeName.STRING),
+                    Field.of("booleanField", StandardSQLTypeName.BOOL),
+                    Field.of("dateField", StandardSQLTypeName.DATE));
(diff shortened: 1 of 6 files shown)


irvifa:
@stephaniewang526 is this okay? I saw error on https://source.cloud.google.com/results/invocations/e7ac9071-fdc8-490a-b6cd-015c38028e55/targets;collapsed=/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-bigquery%2Fpresubmit%2Fjava11-samples/log for only java 11

stephaniewang526:
> @stephaniewang526 is this okay? I saw error on https://source.cloud.google.com/results/invocations/e7ac9071-fdc8-490a-b6cd-015c38028e55/targets;collapsed=/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-bigquery%2Fpresubmit%2Fjava11-samples/log for only java 11

Thanks -- looks like it was a flaky test. We should probs fix that too.

irvifa:
@stephaniewang526 is it specifically only for that test? Because I can see that the BQ table can’t be found when the test was run.

stephaniewang526:
Yes it's specific to the test -- maybe we need to add UUID to tablename. Sometimes we run tests concurrently and it can cause issues. The same test did not fail in Java 8.

PR #415, opened 2020-06-04
docs(samples): rename extract table json to extract table csv

ejdarrow:
@stephaniewang526


ejdarrow:
In Python, bigquery_extract_table is csv, and bigquery_extract_table_json is json.

PR #424, opened 2020-06-08
docs(samples): add update table using dml query sample

### LINKED ISSUE #413: Can you help me the JAVA API to update the table rows using google cloud bigquery
Hi Team,

We are doing the POC with the new Bigquery API https://github.com/googleapis/java-bigquery

Can you please help me how I can update the table rows using this.

**Env:**
Java 1.8 
Google Cloud Bigquery

Thanks,
Shruti

stephaniewang526:
Fixes #413

### DIFF
--- a/samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
+++ b/samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
@@ -34,15 +34,19 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class UpdateTableDMLIT {
+
+  private String tableName;
   private ByteArrayOutputStream bout;
   private PrintStream out;
 
-  private static final String BIGQUERY_DATASET_NAME = System.getenv("BIGQUERY_DATASET_NAME");
+  private static final String BIGQUERY_DATASET_NAME = requireEnvVar("BIGQUERY_DATASET_NAME");
 
-  private static void requireEnvVar(String varName) {
+  private static String requireEnvVar(String varName) {
+    String value =  System.getenv(varName);
     assertNotNull(
         "Environment variable " + varName + " is required to perform these tests.",
         System.getenv(varName));
+    return value;
   }
 
   @BeforeClass
@@ -55,16 +59,9 @@ public class UpdateTableDMLIT {
     bout = new ByteArrayOutputStream();
     out = new PrintStream(bout);
     System.setOut(out);
-  }
 
-  @After
-  public void tearDown() {
-    System.setOut(null);
-  }
-
-  @Test
-  public void testUpdateTableDML() throws IOException, InterruptedException {
-    String tableName = "UserSessions_TEST_" + UUID.randomUUID().toString().replace('-', '_');
+    // Create a test table
+    tableName = "UserSessions_TEST_" + UUID.randomUUID().toString().replace('-', '_');
     Schema schema =
         Schema.of(
             Field.of("id", LegacySQLTypeName.STRING),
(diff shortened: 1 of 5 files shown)


samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class UpdateTableDMLIT {
+  private ByteArrayOutputStream bout;
+  private PrintStream out;
+
+  private static final String BIGQUERY_DATASET_NAME = System.getenv("BIGQUERY_DATASET_NAME");
```
kurtisvg, line 40: You are using `getenv` here, but also have requireEnvVar below. Maybe use `requireEnvVar` here but have it to return a value?

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
```diff
@@ -0,0 +1,85 @@
...
+    String tableName = "UserSessions_TEST_" + UUID.randomUUID().toString().replace('-', '_');
+    Schema schema =
+        Schema.of(
+            Field.of("id", LegacySQLTypeName.STRING),
+            Field.of("user_id", LegacySQLTypeName.STRING),
+            Field.of("login_time", LegacySQLTypeName.STRING),
+            Field.of("logout_time", LegacySQLTypeName.STRING),
+            Field.of("ip_address", LegacySQLTypeName.STRING));
+
+    CreateTable.createTable(BIGQUERY_DATASET_NAME, tableName, schema);
```
kurtisvg, line 76: Should this be set up?

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
```diff
@@ -0,0 +1,85 @@
...
+
+    UpdateTableDML.updateTableDML(BIGQUERY_DATASET_NAME, tableName);
+
+    assertThat(bout.toString()).contains("Table updated successfully using DML");
+
+    // Clean up
+    DeleteTable.deleteTable(BIGQUERY_DATASET_NAME, tableName);
```
kurtisvg, line 83: Should this be teardown?

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
stephaniewang526: We have always done this in the test method in all the other similar samples

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
stephaniewang526: this one also we have been doing this in the method itself...

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
stephaniewang526: this is how we have been handling this in all the samples in BigQuery... is there a new standard?

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
kurtisvg: So this one isn't a big deal, but logically it does make sense that "setUp" call since it's not directly related to a test. 

(consider this one a nit)

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
kurtisvg: This one is more of a concern - if any exception is thrown before this this step won't be executed. This means failures will leak resources.

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
kurtisvg: You're right, this seems to be the same as being used in other places. However, I think maybe this pattern might be improved if we did something like this instead;

```java
  private static final String BIGQUERY_DATASET_NAME = requireEnvVar("BIGQUERY_DATASET_NAME");

  private static string requireEnvVar(String varName) {
   String value =  System.getenv(varName);
    assertNotNull(
        "Environment variable " + varName + " is required to perform these tests.",
        value);
     return value;
  }
```

This way we only have to specify "BIGQUERY_DATASET_NAME" in one pace. 

(consider this one a nit)

stephaniewang526: okay makes sense - done

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
stephaniewang526: done

samples/snippets/src/test/java/com/example/bigquery/UpdateTableDMLIT.java
stephaniewang526: done

PR #456, opened 2020-06-16
samples: migrate samples from GoogleCloudPlatform/java-docs-samples /bigquery/cloud-client

chingor13:


stephaniewang526:
We already migrated!

PR #470, opened 2020-06-19
docs(samples): add create job

pmakani:

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
@@ -48,7 +48,8 @@ public class CreateJob {
               .setLabels(ImmutableMap.of("example-label", "example-value"))
               .build();
 
-      // The location and JobName must be specified; other fields can be auto-detected.
+      // The location and job name are optional,
+      // if both are not specified then client will auto-create.
       String jobName = "jobId_" + UUID.randomUUID().toString();
       JobId jobId = JobId.newBuilder().setLocation("us").setJob(jobName).build();
 
--- a/samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.bigquery;
+
+// [START bigquery_create_job]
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryException;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.Job;
+import com.google.cloud.bigquery.JobId;
+import com.google.cloud.bigquery.JobInfo;
+import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.common.collect.ImmutableMap;
+import java.util.UUID;
+
```
stephaniewang526, line 29: please add description here

samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
```diff
@@ -0,0 +1,70 @@
...
+public class CreateJob {
+
+  public static void runCreateJob() {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String datasetName = "MY_DATASET_NAME";
+    String tableName = "MY_TABLE_NAME";
+    // i.e. SELECT country_name from `bigquery-public-data.utility_us.country_code_iso`
+    String query =
+        "SELECT country_name from `" + projectId + "." + datasetName + "." + tableName + "`";
```
stephaniewang526, line 39: i think we can just use the raw string like you commented out here so we don't have to create all these projectId, datasetName, tableName local vars

samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
```diff
@@ -0,0 +1,70 @@
...
+      QueryJobConfiguration queryConfig =
+          QueryJobConfiguration.newBuilder(query)
+              .setLabels(ImmutableMap.of("example-label", "example-value"))
+              .build();
+
+      // The client libraries automatically generate a job ID.
+      // Override the generated ID with either the job_id_prefix or job_id parameters.
+      String jobId = "code_sample_" + UUID.randomUUID().toString().substring(0, 8);
+      Job job = bigquery.create(JobInfo.of(JobId.of(jobId), queryConfig));
+      job = job.waitFor();
```
stephaniewang526, line 59: could you follow the job format similar to here https://github.com/googleapis/java-bigquery/blob/0ffbac349cc27ce72b06e05723e45a973b69cef1/samples/snippets/src/main/java/com/example/bigquery/LoadLocalFile.java#L70?

samples/snippets/src/main/java/com/example/bigquery/CreateJob.java
```diff
@@ -52,17 +48,21 @@ public static void createJob(String query) {
...
-      String jobId = "code_sample_" + UUID.randomUUID().toString().substring(0, 8);
-      Job job = bigquery.create(JobInfo.of(JobId.of(jobId), queryConfig));
-      job = job.waitFor();
-      if (job.isDone()) {
+      // The location and JobName must be specified; other fields can be auto-detected.
+      String jobName = "jobId_" + UUID.randomUUID().toString();
+      JobId jobId = JobId.newBuilder().setLocation("us").setJob(jobName).build();
+
+      // Create a job with job ID
+      bigquery.create(JobInfo.of(jobId, queryConfig));
```
stephaniewang526, line 56: i believe jobId is optional right? we don't need to create our own. So I think we can link user to the documentation that explains this.

PR #545, opened 2020-07-14
Set policy tags to empty instead of null

tabdulradi:
Currently when BigQuery returns policy tag object without names, the whole policy tag field is set to null, which is not consistent with the REST Api, this pull request changes the behaviour to return an instance with names set to null instead.

This change modifies a previous PR https://github.com/googleapis/java-bigquery/pull/522, that was created in response to my GCP support ticket, slightly tweaking the behaviour to make it more consistent with the REST api.

- [ ] ~Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-bigquery/issues/new/choose) before writing your code!~ There is GCP issue instead. The case Id is 24194276
- [ ] ~Appropriate docs were updated~ (if necessary)

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/PolicyTags.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/PolicyTags.java
@@ -60,10 +60,11 @@ public abstract class PolicyTags implements Serializable {
 
   static PolicyTags fromPb(
       com.google.api.services.bigquery.model.TableFieldSchema.PolicyTags tagPb) {
-    // Treat a PolicyTag message without a Names subfield as invalid.
+    Builder builder = newBuilder();
+    // Treat a PolicyTag message without a Names subfield as empty.
     if (tagPb.getNames() != null) {
-      return newBuilder().setNames(tagPb.getNames()).build();
+      builder = builder.setNames(tagPb.getNames());
     }
-    return null;
+    return builder.build();
   }
 }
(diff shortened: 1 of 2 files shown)


tabdulradi:
@googlebot I signed it!

shollyman:
Thanks for the submission.  However, the backend service response is also being corrected so that it doesn't provide the empty message for policyTags, which is the behavior the first PR mirrors.  In this case, I'm going to recommend that we don't move forward with this PR.

stephaniewang526:
closing per @shollyman comment

marengaz:
@shollyman @stephaniewang526 

the reason for suggesting this change was so that we can tell the difference between the following scenarios:
- policy tag was never added to this field `"policyTags":null`
- policy tag was added to this field `"policyTags":{"names":["projects/myproj/locations/eu/taxonomies/123/policyTags/456"]}`
- policy tag was added to this field, then removed `"policyTags":{"names":[]}`

it mirrors the behaviour of the description field:
- description was never added to this field `"description":null`
- description was added to this field `"description":"my description"`
- description was added to this field, then removed `"description":""`

we value this granularity. whats the reason for removing it?

shollyman:
If you want awareness of this kind of change, then I'd suggest something like the [BigQuery audit log](https://cloud.google.com/bigquery/docs/reference/auditlogs) feature.  If it doesn't convey the information you need, then I'd suggest making a request on the [public issue tracker](https://issuetracker.google.com/savedsearches/559654) or via support channels.

PR #558, opened 2020-07-16
docs(samples): add create view ddl

pmakani:

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/DDLCreateView.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/DDLCreateView.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2020 Google LLC
+ *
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/bigquery/DDLCreateView.java
```diff
@@ -0,0 +1,78 @@
...
+
+// Sample to create a view using DDL
+public class DDLCreateView {
+
+  public static void runDDLCreateView() {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String datasetId = "MY_DATASET_ID";
+    String tableId = "MY_VIEW_ID";
+    String sql =
```
stephaniewang526, line 35: maybe we should name this string `ddl`? since this is a ddl create view sample?

PR #618, opened 2020-07-29
docs(samples): add update routine

pmakani:

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/UpdateRoutine.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/UpdateRoutine.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/bigquery/UpdateRoutine.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  public static void updateRoutine(String datasetName, String routineName) {
+    try {
+      // Initialize client that will be used to send requests. This client only needs to be created
+      // once, and can be reused for multiple requests.
+      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+      Routine routine = bigquery.getRoutine(RoutineId.of(datasetName, routineName));
+      routine.toBuilder().setBody("x * 4").build().update();
```
stephaniewang526, line 43: could we setBody to something else? I don't really understand what the string here means... maybe something straightforward such as "new body text"?

stephaniewang526:
some ITs are failing

samples/snippets/src/main/java/com/example/bigquery/UpdateRoutine.java
pmakani: @stephaniewang526  retrieve the existing routine and updating the body value into a routine , you could check below reference code.

https://github.com/googleapis/java-bigquery/blob/ae97b5fec36daed60a4f51d1e5bdd3438aa079aa/samples/snippets/src/main/java/com/example/bigquery/CreateRoutine.java#L48-L59

PR #653, opened 2020-08-06
feat: support for Table ACLs

shollyman:
This PR adds support for table-level permissions to BigQuery, which is implemented via the IAM meta API.
There's three mechanisms exposed: getIamPolicy, setIamPolicy, and testIamPermissions.

The service exposes a service-centric version of the responses, so we include a policy converter to map to/from the more general cloud IAM messages.

Differences from the other veneer (java-storage):
* changed the response format of TestIamPermissions.  The general pattern is to pass in a resource (table) and a list of permissions to check, and it returns the subset of that permission list which the caller actually holds on the resource.  The storage veneer chose to implement this as a list of bools instead, which is nonstandard.

* Policy conversion is much more forgiving.  In java-storage, it's only top-level resources which return policies, so there's an expectation that all responses have actual bindings.  Tables in BQ are nested within other elements that also convey permissions, so it's far more common and likely to have empty bindings on policies.


pmakani (approved):
LGTM!, except clir and codecov.

PR #714, opened 2020-08-31
chore: update google-cloud-bigquery-parent version in benchmark pom during release

stephaniewang526:
Caused by: #710 and #711 (after google-cloud-bigquery release)

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <artifactId>google-cloud-bigquery-parent</artifactId>
     <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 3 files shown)


benchmark/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <artifactId>google-cloud-bigquery-parent</artifactId>
     <groupId>com.google.cloud</groupId>
-    <version>1.117.1</version>
+    <version>1.117.1</version><!-- {x-version-update:google-cloud-bigquery:current} -->
```
chingor13, line 9: note: current will include snapshot versions

PR #732, opened 2020-09-15
fix: throw jobexception for invalid multiple statements query

stephaniewang526:
Fixes b/163347694


shollyman (approved):
Could you expand on what's going on here in the PR description?  I see you're adding a new raise to the reload, is this to match behavior from something else?

stephaniewang526:
> Could you expand on what's going on here in the PR description? I see you're adding a new raise to the reload, is this to match behavior from something else?

Indeed! Customer noticed that we are throwing Exception in invalid single-statement scenario but not invalid Multi-statement scenario. We are trying to make the behavior consistent here.

PR #737, opened 2020-09-16
feat: add additional time partitioning intervals

stephaniewang526:


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
```diff
@@ -37,16 +37,20 @@
   private static final long serialVersionUID = -8565064035346940951L;
 
   /**
-   * The supported types are DAY, which will generate one partition per day, and HOUR, which will
-   * generate one partition per hour. (Providing an empty string used to cause an error, but in
-   * OnePlatform the field will be treated as unset.)
+   * The supported types are DAY, HOUR, MONTH, and YEAR which will generate one partition per time
+   * unit type. (Providing an empty string used to cause an error, but in OnePlatform the field will
```
shollyman, line 41: Please update the parenthetical phrase.  It's not yet published in docs, but this will be the updated docstring in API ref:
```
[Optional] The supported types are DAY, HOUR, MONTH, and YEAR, which will
generate one partition per day, hour, month, and year, respectively. When
the interval is not specified, the default behavior is DAY.
```

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
```diff
@@ -37,16 +37,20 @@
...
+   * be treated as unset.)
    */
   public enum Type {
 
     /** Table is partitioned per day, based on data loading time. */
```
shollyman, line 46: the "based on data loading time" clause is only true for cases where timepartitioning is not based on user data columns.  Remove it from the various enum comments here.

shollyman (approved):
minor docstring updates

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
stephaniewang526: done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
stephaniewang526: done

pmakani (approved):

PR #776, opened 2020-10-01
feat: add connectionId support for external data sources

stephaniewang526:
Add connecitonId to ExternalTableDefinition to allow external data source query through the BigQuery client.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
@@ -125,7 +125,7 @@ public abstract class ExternalTableDefinition extends TableDefinition {
     public abstract Builder setCompression(String compression);
 
     /**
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
```diff
@@ -124,6 +124,12 @@ public Builder setFormatOptions(FormatOptions formatOptions) {
      */
     public abstract Builder setCompression(String compression);
 
+    /**
+     * [Optional, Trusted Tester] Connection for external data source. The value may be {@code
```
shollyman, line 128: Consider documenting the format of the connectionId string in the comments.

shollyman (approved):

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
stephaniewang526: done. thank you!

PR #835, opened 2020-10-20
docs(samples): create and query Amazon s3 data using external table

stephaniewang526:


samples/snippets/src/main/java/com/example/bigquery/QueryExternalTableAws.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public static void main(String[] args) {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String datasetName = "MY_DATASET_NAME";
+    String externalTableName = "MY_EXTERNAL_TABLE_NAME";
+    String query =
+        String.format(
+            "SELECT * FROM s%.%s.%s WHERE name LIKE 'W%%'",
+            projectId, datasetName, externalTableName);
```
lesv, line 45: Should you explain what this SQL statement does?

samples/snippets/src/main/java/com/example/bigquery/QueryExternalTableAws.java
```diff
@@ -0,0 +1,68 @@
...
+      TableResult results = bigquery.query(QueryJobConfiguration.of(query));
+
+      results
+          .iterateAll()
+          .forEach(row -> row.forEach(val -> System.out.printf("%s,", val.toString())));
+
+      System.out.println("Query on aws external permanent table performed successfully.");
+    } catch (BigQueryException | InterruptedException e) {
+      System.out.println("Query not performed \n" + e.toString());
+    }
```
lesv, line 65: Should these be caught?  Perhaps just mention them or let them propagate?  NOTE - BigQueryException is a RuntimeException and InterruptedException is not.

lesv: More important is why might these have been thrown and what should a user really do with them?

samples/snippets/src/test/java/com/example/bigquery/CreateExternalTableAwsIT.java
```diff
@@ -120,18 +90,19 @@ public void tearDown() throws IOException {
 
   @Test
   public void testCreateExternalTableAws() {
-    String sourceUri = "s3://cloud-samples-tests/us-states.csv";
+    String sourceUri = "s3://steph-omni-test-bucket/us-states.csv";
```
lesv, line 93: Might not be your best final choice for a bucket name.

lesv (approved):
Consider the questions

samples/snippets/src/test/java/com/example/bigquery/CreateExternalTableAwsIT.java
stephaniewang526: Yes - you're right. Created a new bucket for devrel sampels testing. Thank you!

samples/snippets/src/main/java/com/example/bigquery/QueryExternalTableAws.java
stephaniewang526: I think we will catch BigQueryException and not InterruptedException. 
Updated!

samples/snippets/src/main/java/com/example/bigquery/QueryExternalTableAws.java
stephaniewang526: Yes - excellent idea! Thank you!
Done!

PR #886, opened 2020-10-29
ci: nightly cleanup cleanup

stephaniewang526:
```
ERROR: (gcloud.auth.activate-service-account) Unable to read file [/tmpfs/src/gfile/service-acct.json]: [Errno 2] No such file or directory: '/tmpfs/src/gfile/service-acct.json'
```


shollyman (approved):
Config goop is fun.

PR #987, opened 2020-12-02
feat: set requestId for fast query path in QueryRequestInfo instead of QueryJobConfiguration

stephaniewang526:
As requested @epavan

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
@@ -35,6 +35,7 @@ final class QueryRequestInfo {
   private final Long maxResults;
   private final String query;
   private final List<QueryParameter> queryParameters;
+  private final String requestId;
   private final Boolean useQueryCache;
   private final Boolean useLegacySql;
   private final String requestId;
@@ -49,6 +50,7 @@ final class QueryRequestInfo {
     this.maxResults = config.getMaxResults();
     this.query = config.getQuery();
     this.queryParameters = config.toPb().getQuery().getQueryParameters();
+    this.requestId = UUID.randomUUID().toString();
     this.useLegacySql = config.useLegacySql();
     this.useQueryCache = config.useQueryCache();
     this.requestId = config.getRequestId();
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
@@ -23,6 +23,7 @@ import com.google.common.base.Objects;
 import com.google.common.collect.Lists;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -1885,9 +1885,8 @@ public void testFastQueryRequestCompleted() throws InterruptedException {
             .setTotalBytesProcessed(42L)
             .setTotalRows(BigInteger.valueOf(1L));
 
-    QueryRequestInfo requestInfo = new QueryRequestInfo(QUERY_JOB_CONFIGURATION_FOR_QUERY);
-
-    when(bigqueryRpcMock.queryRpc(PROJECT, requestInfo.toPb())).thenReturn(queryResponsePb);
+    when(bigqueryRpcMock.queryRpc(eq(PROJECT), requestPbCapture.capture()))
```
epavan123, line 1888: Need to validate that what is captured is actually what we expect to see (except the requestId).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
```diff
@@ -49,9 +50,9 @@
     this.maxResults = config.getMaxResults();
     this.query = config.getQuery();
     this.queryParameters = config.toPb().getQuery().getQueryParameters();
+    this.requestId = UUID.randomUUID().toString();
```
shollyman, line 53: Are there any cases where a QueryRequestInfo may get reused? Such usage may indicate we set this at time of sending the request (outside of the retry wrapper) rather than time of instantiating the request.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
stephaniewang526: Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
stephaniewang526: Yes -- we decided to design it to set this requestId at the time of instantiating the request object which means when a user runs the same QueryJobConfiguration twice, they will be sending two requestIds.

shollyman: Thanks for clarifying.  My worry was mostly around cases where someone may retain a reference and use this in unexpected ways.

shollyman (approved):

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
stephaniewang526: Noted -- thank you!

epavan123 (approved):

PR #1208, opened 2021-03-30
docs(samples): update querypagination sample

### LINKED ISSUE #1207: Update QueryPagination.java code to better match the comments

**Is your feature request related to a problem? Please describe.**
With regards to the [query pagination example](https://github.com/googleapis/java-bigquery/blob/master/samples/snippets/src/main/java/com/example/bigquery/QueryPagination.java), the current code makes it seem like lines 60-63 would only print the first 20 rows, but it actually prints all of the results since `.iterateAll() ` fetches the next page automatically upon exhausting elements on the given page (mentioned in the [docs](https://googleapis.dev/java/gax/latest/com/google/api/gax/paging/Page.html#iterateAll--)). 

**Describe the solution you'd like**
We can replace `.iterateAll()` with `.getValues()` to only get the results of a single page instead. I've put a PR together with these changes.

**Describe alternatives you've considered**
I tried to use the results of a query job instead of writing to a table and then using the `listTableData` method but the results are the same since `.iterateAll()` does the same thing.

**Additional context**
This caused some confusion for a GCP customer on a support case and I've opened a ticket internally for some other public documentation changes.

ShyamShah11:
Based on feedback from a support case, I've made some changes to make this more clear and more accurate to what the comments indicate

Created https://github.com/googleapis/java-bigquery/issues/1207 to track this

Fixes #1207 ☕️

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/QueryPagination.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/QueryPagination.java
@@ -59,15 +59,23 @@ public class QueryPagination {
 
       // First Page
       results
-          .iterateAll()
-          .forEach(row -> row.forEach(val -> System.out.printf("%s,", val.toString())));
+          .getValues()
+          .forEach(row -> row.forEach(val -> System.out.printf("%s,\n", val.toString())));
 
       if (results.hasNextPage()) {
         // Next Page
         results
             .getNextPage()
+            .getValues()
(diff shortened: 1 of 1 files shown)


stephaniewang526:
Could you sign the cla?

Also, could you update your initial commit as well as the subject line of this PR to: `docs(samples): update querypagination sample`

stephaniewang526 (changes requested):
Could you sign the cla?

Also, could you update your initial commit as well as the subject line of this PR to: `docs(samples): update querypagination sample`.

Issue #1244, opened 2021-04-16
minor: integration test for column ACLs feature


shollyman:
Now that datacatalog v1 endpoint has support for the PolicyTagManager client, let's add an integration test to exercise the feature across the service boundary.  More details  shared out of band with implementors.

PR #1266, opened 2021-04-29
test: add integration test for Column ACLs

stephaniewang526:
Fixes: #1244

### DIFF
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -31,6 +31,10 @@
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-datacatalog</artifactId>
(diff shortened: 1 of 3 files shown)


shollyman (approved):
Thanks for putting this together.  Not sure I understand the changes in dependencies.sh, but you're far more familiar with java dependency management than I am.

stephaniewang526:
> Thanks for putting this together. Not sure I understand the changes in dependencies.sh, but you're far more familiar with java dependency management than I am.

Thanks! Yes, I reported this [known issue with maven-dependency-plugin](https://issues.apache.org/jira/browse/MDEP-737) on 03/01/2021 and haven't heard anything back yet. Therefore, adding the exclusion as a temporary fix. I added this issue in the comment in case it gets fixed so we can remove it.

PR #1282, opened 2021-05-05
feat: configure initial sp version

stephaniewang526:


chingor13 (approved):

Neenu1995 (approved):

shollyman (commented):
I have no context on this, and I'm unable to infer what "sp" refers to in this case.  The use of null for pull request matching seems odd; should it simply be elided?

stephaniewang526:
> I have no context on this, and I'm unable to infer what "sp" refers to in this case. The use of null for pull request matching seems odd; should it simply be elided?

this is related to b/187189809

Issue #1300, opened 2021-05-11
MaterializedViewDefinition with partitioning and clustering


jessicalu2016:
Hi guys,

is there a way to specify partitioning and clustering when creating a Materialized View? `MaterializedViewDefinition` doesn't seem to provide a way to do it.

Thanks,
Jessica

stephaniewang526:
Hi, this feature is supported in v1.130.0. Please try it out and re-open the ticket/open a new ticket if there are other issues.

jessicalu2016:
awesome, Thanks!

PR #1301, opened 2021-05-11
feat: add support for partitioning and clustering in MaterializedViewDefinition

stephaniewang526:
Towards #1300

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/MaterializedViewDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/MaterializedViewDefinition.java
@@ -58,20 +58,20 @@ public abstract class MaterializedViewDefinition extends TableDefinition {
     public abstract Builder setType(Type type);
 
     /**
-     * Sets the time partitioning configuration for the table. If not set, the table is not
+     * Sets the time partitioning configuration for the materialized view. If not set, the materialized view is not
      * time-partitioned.
      */
     public abstract Builder setTimePartitioning(TimePartitioning timePartitioning);
 
     /**
-     * Sets the range partitioning configuration for the table. Only one of timePartitioning and
+     * Sets the range partitioning configuration for the materialized view. Only one of timePartitioning and
      * rangePartitioning should be specified.
      */
     public abstract Builder setRangePartitioning(RangePartitioning rangePartitioning);
 
     /**
-     * Set the clustering configuration for the table. If not set, the table is not clustered.
-     * BigQuery supports clustering for both partitioned and non-partitioned tables.
+     * Set the clustering configuration for the materialized view. If not set, the materialized view is not clustered.
+     * BigQuery supports clustering for both partitioned and non-partitioned materialized views.
      */
     public abstract Builder setClustering(Clustering clustering);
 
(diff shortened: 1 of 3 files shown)


shollyman (commented):
Are there other things that should get hoisted here?  Reviewing https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table and the existing StandardTableDefinition.java, I suspect we want to also pull in stats messages like numrows/numbytes/location.  It doesn't appear like java ever hoisted requirePartitionFilter out of the partitioning messages, so that's something to consider as a future FR.

stephaniewang526:
> Are there other things that should get hoisted here? Reviewing https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table and the existing StandardTableDefinition.java, I suspect we want to also pull in stats messages like numrows/numbytes/location. It doesn't appear like java ever hoisted requirePartitionFilter out of the partitioning messages, so that's something to consider as a future FR.

I think we should probs evaluate the delta and add those in in a separate PR.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/MaterializedViewDefinition.java
```diff
@@ -57,6 +57,24 @@
     @Override
     public abstract Builder setType(Type type);
 
+    /**
+     * Sets the time partitioning configuration for the table. If not set, the table is not
```
shollyman, line 61: nit: do we want to update the comments here and elsewhere to indicate materialized view, rather than table?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/MaterializedViewDefinition.java
```diff
@@ -149,6 +197,28 @@ static MaterializedViewDefinition fromPb(Table tablePb) {
       if (materializedViewDefinition.getRefreshIntervalMs() != null) {
         builder.setRefreshIntervalMs(materializedViewDefinition.getRefreshIntervalMs());
       }
+      if (tablePb.getTimePartitioning() != null) {
+        try {
+          builder.setTimePartitioning(TimePartitioning.fromPb(tablePb.getTimePartitioning()));
+        } catch (IllegalArgumentException e) {
+          throw new IllegalArgumentException(
+              "Illegal Argument - Got unexpected time partitioning "
+                  + tablePb.getTimePartitioning().getType()
```
shollyman, line 206: Do you want to intercept like this?  Seems like we may want the other fromPb() message functions to be responsible for throwing a human readable error?

shollyman (approved):
Just a couple minor comments for consideration.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/MaterializedViewDefinition.java
stephaniewang526: Yes -- Updated to `materialized view`.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/MaterializedViewDefinition.java
stephaniewang526: Okay - I removed this interception.

PR #1390, opened 2021-06-24
feat: add support for parameterized type

### LINKED ISSUE #1309: FR: add support for parameterized types
This FR tracks the impl for plumbing through newly supported type constraints in table schemas.  STRING and BYTES have maxlength, NUMERIC/BIGNUMERIC support constraints on precision and scale.

Additional guidance shared out of band with library maintainers.

stephaniewang526:
Fixes #1309

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
@@ -293,17 +293,20 @@ public final class Field implements Serializable {
     return policyTags;
   }
 
-  /** Returns the max length for the field. */
+  /** Returns the maximum length of the field for STRING or BYTES type. */
   public Long getMaxLength() {
     return maxLength;
   }
 
-  /** Returns the scale for the field. */
+  /**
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
```diff
@@ -244,6 +293,21 @@ public PolicyTags getPolicyTags() {
     return policyTags;
   }
 
+  /** Returns the max length for the field. */
```
shollyman, line 296: Maybe augment the javadocs for the public getters with the information about what types each of these relates to, but no need to be as exhaustive as the builder's setters.

shollyman (approved):
Minor nit on javadoc, but the PR itself looks good

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
stephaniewang526: Makes sense. Update Javadoc.

PR #1409, opened 2021-06-29
samples: use a uuid for insert id in streaming insert with row id example

anguillanneuf:
IIUC, the first streaming insert Java sample [here](https://cloud.google.com/bigquery/streaming-data-into-bigquery#streaminginsertexamples) is supposed to show streaming inserts with insert IDs, which the Java client calls `id` and does not populate using UUID if unset.

- Go calls it `insertId` and autopopulates it: https://pkg.go.dev/cloud.google.com/go/bigquery#ValuesSaver
- Python called it `row_id` and under the hood implements it via a `insertId` if unset https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/client.py#L3423

---
Also see internal cl/382167079

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/TableInsertRows.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/TableInsertRows.java
@@ -27,6 +27,7 @@ import com.google.cloud.bigquery.TableId;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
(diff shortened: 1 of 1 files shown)


stephaniewang526:
Thanks for making this change but we don't want to show this and ideally other languages should remove this from their samples too.

anguillanneuf:
I see now - Java matches Go sample, but the Python sample is doing something else because of a different default implementation. Thanks for the review!

PR #1431, opened 2021-07-08
feat: add dmlStatistics support

stephaniewang526:
Provides detailed statistics for DML statements Present only for DML statements INSERT, UPDATE, DELETE or TRUNCATE.

Towards b/186432630

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
@@ -625,9 +625,7 @@ public abstract class JobStatistics implements Serializable {
       return numDmlAffectedRows;
     }
 
-    /**
-     * Detailed statistics for DML statements.
-     */
+    /** Detailed statistics for DML statements. */
     public DmlStats getDmlStats() {
       return dmlStats;
     }
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
@@ -626,8 +626,7 @@ public abstract class JobStatistics implements Serializable {
     }
 
     /**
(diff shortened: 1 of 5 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
```diff
@@ -614,6 +625,14 @@ public Long getNumDmlAffectedRows() {
       return numDmlAffectedRows;
     }
 
+    /**
+     * Detailed statistics for DML statements Present only for DML statements INSERT, UPDATE, DELETE
```
shollyman, line 629: Consider being less specific about which statements may yield stats?  There's at least two DML statement types missing here, MERGE and TRUNCATE TABLE.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -2205,6 +2205,26 @@ public void testFastQueryHTTPException() throws InterruptedException {
     }
   }
 
+  @Test
+  public void testDmlStatistics() throws InterruptedException {
+    String tableName = TABLE_ID_FASTQUERY.getTable();
```
shollyman, line 2210: Only concern here is whether this UPDATE will impact other tests that use this table, as test execution order is not guaranteed.

shollyman (approved):

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
stephaniewang526: I think this should be okay since the WHERE is set to TRUE so it will always just update the table (2 rows).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
stephaniewang526: Okay -- reduced comment here.

PR #1497, opened 2021-08-10
feat: add support for transactioninfo in query statistics

### LINKED ISSUE #1467: bigquery: support transactioninfo in query statistics
This FR tracks support for the transactionInfo in query statistics, which enables users to correlated jobs to a specific transaction.  Details about this change are shared out of band with library maintainers.

stephaniewang526:
Fixes #1467

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
@@ -2222,21 +2222,13 @@ public class ITBigQueryTest {
 
   @Test
   public void testTransactionInfo() throws InterruptedException {
+    String tableName = TABLE_ID_FASTQUERY.getTable();
     String transaction =
-        "BEGIN\n"
-            + "\n"
-            + "  BEGIN TRANSACTION;\n"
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -2219,6 +2220,27 @@ public void testDmlStatistics() throws InterruptedException {
     assertEquals(2L, statistics.getDmlStats().getUpdatedRowCount().longValue());
   }
 
+  @Test
+  public void testTransactionInfo() throws InterruptedException {
+    String tableName = TABLE_ID_FASTQUERY.getTable();
```
shollyman, line 2225: Do you need to worry about test ordering, or is this created in a setup?

shollyman (approved):

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
stephaniewang526: Thank you for pointing this out. Yes, this is fine since the later test case will still update the table given the `WHERE TRUE` condition.

PR #1512, opened 2021-08-21
feat: Adding logic for Apache Arrow vs Avro Benchmark 

prash-mi:
The code benchmarks Apache Arrow vs Avro while de-serializing the records to the row level.
Sample output: 

```
Benchmark                                  Mode  Cnt   Score   Error  Units
BQArrowBenchMark.deserializationBenchmark  avgt    5  12.442 ± 0.252  ms/op
BQAvroBenchMark.deserializationBenchmark   avgt    5  28.656 ± 0.443  ms/op
```

### DIFF
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -15,6 +15,7 @@
   </parent>
   <properties>
     <site.installationModule>google-cloud-bigquery</site.installationModule>
+    <jmh.version>1.32</jmh.version>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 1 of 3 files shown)


suztomo:
@stephaniewang526  Apache Beam recently started using Apache Arrow. https://github.com/apache/beam/tree/master/sdks/java/extensions/arrow
I'm not sure how Apache Arrow's different versions would be problematic (it was certainly problematic in case of Avro) but please pay attention to the version discrepancy. (No version discrepancy is desirable.) I just hope Apache Arrow versions wouldn't cause problems for Dataflow users who also use BigQuery.

stephaniewang526:
> @stephaniewang526 Apache Beam recently started using Apache Arrow. https://github.com/apache/beam/tree/master/sdks/java/extensions/arrow
> I'm not sure how Apache Arrow's different versions would be problematic (it was certainly problematic in case of Avro) but please pay attention to the version discrepancy. (No version discrepancy is desirable.) I don't advise which version of Apache Arrow BigQuery and Beam should use (there's [an open PR in Beam to bump the version from 4 to 5](https://github.com/apache/beam/pull/15359).). I just hope Apache Arrow version discrepancy, if any, wouldn't cause problems for Dataflow users who also use BigQuery.

Hi Tomo, yes this is a good point. I don't mind using the version of Arrow that Beam is using if you could let us know what that is. There are some known issues w Arrow dependency management similar to Avro, unfortunately. We are leaning towards using Arrow in the BigQuery project.

stephaniewang526:
@prash-mi could you update the initial commit to `chore: add benchmark for arrow vs. avro` so that this change will not trigger a minor bump in the release version of the client library?

PR #1628, opened 2021-09-30
fix: BigQueryRetryAlgorithm using a static for uuid

BenWhitehead:
* update log lines to be on a single line rather than multi-line
* add test to verify running multiple things via BigQueryRetryHelper#runWithRetry

### DIFF
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -76,6 +76,10 @@
       <groupId>com.google.api</groupId>
       <artifactId>api-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client</artifactId>
+    </dependency>
 
     <dependency>
       <groupId>com.google.guava</groupId>
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -76,10 +76,6 @@
       <groupId>com.google.api</groupId>
       <artifactId>api-common</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.google.api-client</groupId>
-      <artifactId>google-api-client</artifactId>
-    </dependency>
 
     <dependency>
       <groupId>com.google.guava</groupId>
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryRetryHelperTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryRetryHelperTest.java
@@ -18,10 +18,6 @@ package com.google.cloud.bigquery;
 
 import static com.google.common.truth.Truth.assertThat;
 
-import com.google.api.client.googleapis.json.GoogleJsonError;
-import com.google.api.client.googleapis.json.GoogleJsonResponseException;
-import com.google.api.client.http.HttpHeaders;
-import com.google.api.client.http.HttpResponseException;
 import com.google.api.core.NanoClock;
 import com.google.api.gax.retrying.BasicResultRetryAlgorithm;
 import com.google.api.gax.retrying.RetrySettings;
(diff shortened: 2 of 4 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryRetryHelperTest.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.junit.Test;
+import org.threeten.bp.Duration;
+
+public class BigQueryRetryHelperTest {
+
+  @Test
+  public void parallelRetriesAllBehaveAsExpected() throws ExecutionException, InterruptedException {
```
stephaniewang526, line 52: I think we should rename this to be something like `testRetriesMultithread` since we may need to add parallel processes ones later.

google-cloud-bigquery/pom.xml
```diff
@@ -76,6 +76,10 @@
       <groupId>com.google.api</groupId>
       <artifactId>api-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client</artifactId>
```
stephaniewang526, line 81: What is this new dependency used for?

stephaniewang526 (changes requested):

google-cloud-bigquery/pom.xml
BenWhitehead: Previously it was a transitive dependency needed for the apiary client. In the test I added, one of the exceptions I throw is a Google Json exception, the same kind apiary clients produce. If I mark that dependency as `<scope>test</scope>` that removes it from the compile classpath and breaks the compile, so I had to add it as a non-test dependency.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryRetryHelperTest.java
BenWhitehead: done

google-cloud-bigquery/pom.xml
stephaniewang526: Can we simplify the test to now use this new dependency?

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryRetryHelperTest.java
```diff
@@ -0,0 +1,170 @@
...
+                .collect(Collectors.toList()));
+
+    List<Result> results = f.get();
+
+    List<String> values = results.stream().map(r -> r.val).collect(Collectors.toList());
+    Map<String, List<String>> grouped =
+        values.stream().collect(Collectors.groupingBy(Function.identity()));
+    int sum = results.stream().mapToInt(r -> r.count).sum();
+
+    assertThat(values).hasSize(endInclusive);
```
stephaniewang526, line 83: Can we simplify this test case to just show we are retrying with the correct maxAttempt and let's make maxAttempt a variable which is being asserted on. 

The idea here is to verify that we are not doing infinite retries and we should show that as clearly as possible.

google-cloud-bigquery/pom.xml
BenWhitehead: changes to using Connect exception instead

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryRetryHelperTest.java
BenWhitehead: I've simplified this a bit and added some comments to explain what is being asserted.

These assertions are trying to ensure all of the retries are happening across all of the jobs and that all jobs ultimately succeed so I think we need to keep at least the two asserts I've left.

stephaniewang526 (approved):

stephaniewang526:
Thank you for the contribution. Closing this PR for now as we are actively working with partner engineers to determine the next steps regarding client retry expectations.

PR #1630, opened 2021-09-30
feat: add support for AvroOptions

stephaniewang526:
New AvroOptions class and wired into ExternalTableDefinition


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FormatOptions.java
```diff
@@ -120,6 +120,8 @@ public static FormatOptions of(String format) {
     checkArgument(!isNullOrEmpty(format), "Provided format is null or empty");
     if (format.equals(CSV)) {
       return csv();
+    } else if (format.equals(AVRO)) {
+      return avro();
```
shollyman, line 124: don't you need to also update the avro() function to use your new AvroOptions type?

stephaniewang526: Yes, good catch. Done. Thanks!

shollyman (approved):

PR #1645, opened 2021-10-08
fix: fix timestamp rounding issue

### LINKED ISSUE #1644: BigQuery: Timestamp rounding issue
#### Environment details

1. BigQuery
2. OS type and version: macOS
3. Java version: 8
4. bigquery version(s): at commit: 2b942da

#### Code example

I run the following query (note the last 3 digits are **574**): 

`SELECT TIMESTAMP '2022-01-24T23:54:25.095574Z'`

but I get back (note the last 3 digits are **573**): 

`2022-01-24T23:54:25.095573Z`

The following is a screen shot of the test and the result that I added to `src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java`:

BigQueryTimestamp

Seems the java client is affected by this issue, If I run the same query in BigQuery web console, the query returns correctly: 
Screen Shot 2021-10-08 at 6 39 28 pm

#### Steps to reproduce

apply this [patch](https://github.com/liu-du/java-bigquery/commit/e207a247b58626bcc0a08509e8a3f7b8790c29c7) and run the "tmpTest" in `src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java`.

#### Any additional information below

- Only certain timestamp values may cause the rounding error
- it seems that query priority has to be set to INTERATIVE to reproduce the rounding error
- Although BigQuery Timestamp is up to microsecond according to documentation, but the value in FieldValue (see screenshot above) actually returns 7 digits instead of 6 (again this only happens for certain timestamps).

the following patch fixes the problem by rounding the big decimal to neareast whole number before taking the longValue:

```diff
diff --git a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
index a2ce187..5bab4b3 100644
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
@@ -25,6 +25,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.io.BaseEncoding;
 import java.io.Serializable;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -183,7 +184,7 @@ public class FieldValue implements Serializable {
     // timestamps are encoded in the format 1408452095.22 where the integer part is seconds since
     // epoch (e.g. 1408452095.22 == 2014-08-19 07:41:35.220 -05:00)
     BigDecimal secondsWithMicro = new BigDecimal(getStringValue());
-    BigDecimal scaled = secondsWithMicro.scaleByPowerOfTen(6);
+    BigDecimal scaled = secondsWithMicro.scaleByPowerOfTen(6).setScale(0, RoundingMode.HALF_UP);
     return scaled.longValue();
   }
```

stephaniewang526:
Towards #1644

Temp fix timestamp rounding error (to microsecond) which can be reproduced by setting query priority to `INTERACTIVE`.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
@@ -184,6 +184,9 @@ public class FieldValue implements Serializable {
     // timestamps are encoded in the format 1408452095.22 where the integer part is seconds since
     // epoch (e.g. 1408452095.22 == 2014-08-19 07:41:35.220 -05:00)
     BigDecimal secondsWithMicro = new BigDecimal(getStringValue());
+    // Rounding the BigDecimal to the nearest whole number before setting the longValue in order to
+    // address TimeStamp rounding issue described in
+    // https://github.com/googleapis/java-bigquery/issues/1644
     BigDecimal scaled = secondsWithMicro.scaleByPowerOfTen(6).setScale(0, RoundingMode.HALF_UP);
     return scaled.longValue();
   }
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
@@ -25,6 +25,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.io.BaseEncoding;
 import java.io.Serializable;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
```diff
@@ -183,7 +184,7 @@ public long getTimestampValue() {
     // timestamps are encoded in the format 1408452095.22 where the integer part is seconds since
     // epoch (e.g. 1408452095.22 == 2014-08-19 07:41:35.220 -05:00)
     BigDecimal secondsWithMicro = new BigDecimal(getStringValue());
-    BigDecimal scaled = secondsWithMicro.scaleByPowerOfTen(6);
+    BigDecimal scaled = secondsWithMicro.scaleByPowerOfTen(6).setScale(0, RoundingMode.HALF_UP);
```
tswast, line 187: Let's add a comment here as to why rounding is needed.

stephaniewang526: Done. 
I will leave the issue open.

tswast (approved):

PR #1650, opened 2021-10-12
debug: add a test case for timestamp issue

stephaniewang526:
Towards b/139860901


stephaniewang526:
Clarified that everything is working as expected. We need to pass `String` instead of `long` for subsecond TimeStamp precision.

PR #1700, opened 2021-11-12
fix: removing a new line character in a property

suztomo:
I found the pom.xml has a strange white space character in pom.xml.

Screen Shot 2021-11-12 at 3 04 12 PM

Maybe Maven just ignores such white space characters but it's better to fix it.


stephaniewang526 (approved):

stephaniewang526:
Thanks for noticing and fixing!

PR #1701, opened 2021-11-12
feat: set Table.Schema for permanent external tables

stephaniewang526:
Per discussion related to go/bq-external-table-schema

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
@@ -954,6 +954,28 @@ public class ITBigQueryTest {
   }
 
   @Test
+  public void testSetPermExternalTableSchema() {
+    String tableName = "test_create_external_table_perm";
+    TableId tableId = TableId.of(DATASET, tableName);
+    ExternalTableDefinition externalTableDefinition =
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -953,6 +953,28 @@ public void testCreateExternalTable() throws InterruptedException {
...
+            .setSchema(TABLE_SCHEMA)
+            .setConnectionId(
+                "projects/java-docs-samples-testing/locations/us/connections/DEVREL_TEST_CONNECTION")
+            .build();
+    TableInfo tableInfo = TableInfo.of(tableId, externalTableDefinition);
+    Table createdTable = bigquery.create(tableInfo);
+
+    assertNotNull(createdTable);
+    assertEquals(DATASET, createdTable.getTableId().getDataset());
+    assertEquals(tableName, createdTable.getTableId().getTable());
```
shollyman, line 972: It seems like the significant change here is where schema is set, but you're not really probing that with this test.

stephaniewang526: without the code change the IT will fail.

shollyman (approved):

PR #1723, opened 2021-12-01
feat: add support for BI Engine Statistics

stephaniewang526:
fixes b/205146044


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -2900,6 +2900,13 @@ public void testQueryJob() throws InterruptedException, TimeoutException {
     assertTrue(bigquery.delete(destinationTable));
     Job queryJob = bigquery.getJob(remoteJob.getJobId());
     JobStatistics.QueryStatistics statistics = queryJob.getStatistics();
+    assertNotNull(statistics.getBiEngineStats());
```
shollyman, line 2903: This means all test projects will need to have a reservation allocated, doesn't it?  Wrapping the other assertions in an if block for when getBiEngineStats != null allows you to test if the reservation is present, but doesn't fail if its not.

stephaniewang526: Yes, good point. Updated code. Thanks!

shollyman (approved):

PR #1744, opened 2021-12-07
feat: create Job retry for rate limit exceeded with status code 200

franklinWhaite:
With these changes Create Job will retry for rate limit exceeded even if the status code is 200


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -347,37 +347,46 @@ public JobId get() {
...
 
+    final Map<BigQueryRpc.Option, ?> optionsMap = optionMap(options);
     BigQueryException createException;
     // NOTE(pongad): This double-try structure is admittedly odd.
     // translateAndThrow itself throws, and pretends to return an exception only
     // so users can pretend to throw.
     // This makes it difficult to translate without throwing.
     // Fixing this entails some work on BaseServiceException.translate.
     // Since that affects a bunch of APIs, we should fix this as a separate change.
+    final JobInfo jobInfoCopy = jobInfo;
```
stephaniewang526, line 360: Please add comment to explain the motivation for doing this.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -347,37 +347,46 @@ public JobId get() {
...
         return Job.fromPb(
             this,
-            runWithRetries(
+            BigQueryRetryHelper.runWithRetries(
                 new Callable<com.google.api.services.bigquery.model.Job>() {
                   @Override
                   public com.google.api.services.bigquery.model.Job call() {
-                    return bigQueryRpc.create(jobPb, optionsMap);
+                    if(idRandom){
+                      // generate new random job
```
stephaniewang526, line 371: Maybe we can expand this comment to say "re-generate a new random job with the same jobInfo when jobId is not provided by the user"

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -347,37 +347,46 @@ public JobId get() {
...
+    final Map<BigQueryRpc.Option, ?> optionsMap = optionMap(options);
     BigQueryException createException;
     // NOTE(pongad): This double-try structure is admittedly odd.
     // translateAndThrow itself throws, and pretends to return an exception only
     // so users can pretend to throw.
     // This makes it difficult to translate without throwing.
     // Fixing this entails some work on BaseServiceException.translate.
     // Since that affects a bunch of APIs, we should fix this as a separate change.
+    final JobInfo jobInfoCopy = jobInfo;
+    final JobId[] finalJobId = new JobId[1];
```
stephaniewang526, line 361: Does this have to be an array? Can we just use `jobId`? since we are only storing a single, final jobId?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -347,37 +347,46 @@ public JobId get() {
...
             this,
-            runWithRetries(
+            BigQueryRetryHelper.runWithRetries(
                 new Callable<com.google.api.services.bigquery.model.Job>() {
                   @Override
                   public com.google.api.services.bigquery.model.Job call() {
-                    return bigQueryRpc.create(jobPb, optionsMap);
+                    if(idRandom){
+                      // generate new random job
+                      JobInfo newJobInfo = jobInfoCopy.toBuilder().setJobId(idProvider.get()).build();
```
stephaniewang526, line 372: Maybe rename this variable to `recreatedJobInfo` since everything but JobId is the same for this Job.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryRetryAlgorithm.java
```diff
@@ -92,13 +92,24 @@ public boolean shouldRetry(
...
+    String errorDesc = null;
+    if (previousThrowable != null ){
+      errorDesc = previousThrowable.getMessage();
+    } else if (previousResponse != null){
+      /*
+      In some cases error messages may come without an exception
+      e.g. status code 200 with a rate limit exceeded for job create
+      in these cases there is now previousThrowable so we need to check previousResponse
+       */
+      errorDesc = previousResponse.toString();
```
stephaniewang526, line 109: Please add a mock test in `BigQueryImplTest` to verify the correctness of the behavior; i.e. retrying with status code 200 on rateLimitExceeded error message.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
franklinWhaite: Actually not needed, removed this variable

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
franklinWhaite: expanded comment as recommended

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
franklinWhaite: changed variable name as recommended

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryRetryAlgorithm.java
franklinWhaite: added mock test to `BigQueryImplTest` test is called `testCreateJobFailureShouldRetry`

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
franklinWhaite: Tried several approaches and using an array is the only viable option i found. My understanding is that the variable needs to be a final in order to be accessed from within the inner class (BigQueryRetryHelper). If it's a final it can not be modified, therefore the new jobId values cannot be assigned to it

I would really appreciate any suggestions on how to avoid using an array here.

stephaniewang526 (approved):

PR #1745, opened 2021-12-08
chore: added the closing tag for dependencies

viveknaskar:
I found that the maven dependency for the google-cloud-bigquery was incomplete while I was revisiting the code. Hence, added the required closing tag to fix the issue.

PR #1749, opened 2021-12-14
chore: update README to add missing closing tag

stephaniewang526:
Related to #1745

### DIFF
--- a/samples/snippets/pom.xml
+++ b/samples/snippets/pom.xml
@@ -100,7 +100,9 @@
       <version>1.1.3</version>
       <scope>test</scope>
     </dependency>
+  <!-- [START bigquery_java_dependencies] -->
   </dependencies>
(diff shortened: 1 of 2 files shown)


samples/snippets/pom.xml
```diff
@@ -100,7 +100,9 @@
       <version>1.1.3</version>
       <scope>test</scope>
     </dependency>
+  <!-- [START bigquery_java_dependencies] -->
```
chingor13, line 103: This isn't the snippet that is used in the README, you want the `bigquery_install_with_bom` one

chingor13 (changes requested):
If it's correct, owlbot post-processor should update the README in this PR.

samples/snippets/pom.xml
stephaniewang526: done. thanks!

chingor13 (approved):

Issue #1747, opened 2021-12-08
Update references to package in `readme.md` with Renovate


HonkingGoose:
**Is your feature request related to a problem? Please describe.**

@stephaniewang526 needed to update the `README.md` manually after a dependency update by Renovate: https://github.com/googleapis/java-bigquery/pull/1735
Pushing new commits to a Renovate branch blocks Renovate from merging automatically or pushing new commits into the `renovate/**` branch.

I think you're better of automating the syncing of the README.md, so you can have 2 separate PRs:

- One PR to update the "real" dependency
- One PR to update the `README.md` dependency reference

_Maybe_ you can even make it so that you just get one PR which includes both the `README.md` and the real dependency update.

**Describe the solution you'd like**

I think you can probably get most of the way there by looking at the Renovate configuration file [^renovate-config] that the Renovate team uses. There's also a real world PR for you to look at to see the behavior. [^pr]

**Describe alternatives you've considered**

Keep doing things manually.

**Additional context**

Read the Renovate documentation for more information on the configuration options. [^docs]

[^docs]: https://docs.renovatebot.com/
[^renovate-config]: https://github.com/renovatebot/.github/blob/1cfabab3839113ac37eacb27605cb26757db5eeb/default.json#L99-L108
[^pr]: https://github.com/renovatebot/renovate/pull/12964

stephaniewang526:
Hi, my question was not related to updated README. It was related to the actual dependency update. Looks like the ticket is deleted? https://github.com/renovatebot/renovate/issues/12994

HonkingGoose:
Your issue was converted to a discussion by the maintainer of Renovate, you can find the response here: https://github.com/renovatebot/renovate/discussions/12997 😉 

I saw that you needed to push changes to the PR branch from Renovate, which blocks Renovate from doing anything else with that branch (to prevent Renovate from deleting your wanted changes!).

But I think you can setup Renovate to update the `README.md` in a separate/grouped PR.
The big idea is that you can use `regexManager` to create your own custom rules, as you can see in the example I gave above.

HonkingGoose:
I see that you assigned me, but I don't know enough to configure this properly. So I'll let you or your team do this. 😉

stephaniewang526:
Okay -- following up here: https://github.com/renovatebot/renovate/discussions/12997#discussioncomment-1807949

stephaniewang526:
Fixed by importing the `google-cloud-data-catalog` BOM.

PR #1751, opened 2021-12-15
chore: import datacatalog BOM to avoid renovate-bot issues

stephaniewang526:
Towards #1747

### DIFF
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -31,14 +31,6 @@
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
-      <artifactId>google-cloud-datacatalog</artifactId>
-    </dependency>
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/pom.xml
```diff
@@ -116,6 +108,16 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-datacatalog</artifactId>
+      <scope>test</scope>
```
chingor13, line 114: These are actually test dependencies?

stephaniewang526: yes -- only used in IT

Neenu1995 (approved):

PR #1754, opened 2021-12-16
docs: rename alter materialized view to update

shollyman:
Alter implies a DDL sql ALTER MATERIALIZED VIEW statement, the code sample is purely an API
update example.


stephaniewang526:
Thank you!

PR #1762, opened 2021-12-30
docs: fix query pagination sample

shollyman:
Sample code was misleading; it ran a query, then read from a table.
It did not, in fact, read the query results.

This has caused confusion for users;
https://stackoverflow.com/questions/70509827/bigquery-client-bigquery-listtabledata-with-filter-condition
is an example where a user is confused by the behavior because they're
starting with broken code in the first place.


shollyman:
Actually, going to close this.  The sample is correct, but a bit awkward.  It saves to a destination table and then uses tabledata.list, because there's an appropriate pagination option exposed there which isn't present as a JobOption.

PR #1763, opened 2021-12-30
feat: add Dataset ACL support

stephaniewang526:
Fixes b/211633697

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
@@ -24,6 +24,7 @@ import com.google.api.services.bigquery.model.DatasetAccessEntry;
 import com.google.cloud.StringEnumType;
 import com.google.cloud.StringEnumValue;
 import java.io.Serializable;
+import java.util.List;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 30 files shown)


stephaniewang526:
Blocked by cl/421673680. Will update the generated client version once the CL is released.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
```diff
@@ -119,6 +122,11 @@ public Type getType() {
     abstract Access toPb();
 
     static Entity fromPb(Access access) {
+      if (access.getDataset() != null) {
+        return new Dataset(
```
shollyman, line 126: minor constructor surprise for me, until I realized we're dealing with an ACL.Dataset and not a Dataset.  I suspect this is just local ambiguity; any other places where we need to consider the naming?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
```diff
@@ -146,6 +154,64 @@ static Entity fromPb(Access access) {
     }
   }
 
+  /**
+   * Class for a BigQuery Dataset entity. Objects of this class represent a dataset from a different
```
shollyman, line 158: Maybe clarify this is a Dataset ACL entity, not a Dataset entity?

shollyman (approved):
Mostly my concern is naming confusion, but the precedent was set before this change made.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
stephaniewang526: Yeah I agree it's a little confusing so I renamed it to be more explicit: `DatasetAclEntity`.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
stephaniewang526: Updated. Thanks!

PR #1771, opened 2022-01-05
docs(bigquery): remove duplicate IAM policy update sample

shollyman:
The "create" sample is actually the update sample, and is functionally
identical to the existing UpdateIamPolicy snippet.


stephaniewang526:
The current samples seem fine to me -- the [API calls for create vs. update](https://cloud.google.com/bigquery/docs/table-access-controls#api) are similar that's why the Java samples are similar.

PR #1799, opened 2022-01-18
feat: add JSON type support

stephaniewang526:
Fixes b/213373577

cc @nageshs

### DIFF
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -81,6 +81,10 @@
       <groupId>org.threeten</groupId>
       <artifactId>threetenbp</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+    </dependency>
 
     <!-- auto-value creates a class that uses an annotation from error_prone_annotations -->
     <dependency>
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -81,6 +81,10 @@
       <groupId>org.threeten</groupId>
       <artifactId>threetenbp</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+    </dependency>
 
     <!-- auto-value creates a class that uses an annotation from error_prone_annotations -->
     <dependency>
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -81,10 +81,6 @@
       <groupId>org.threeten</groupId>
       <artifactId>threetenbp</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.json</groupId>
-      <artifactId>json</artifactId>
-    </dependency>
 
     <!-- auto-value creates a class that uses an annotation from error_prone_annotations -->
     <dependency>
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
@@ -61,6 +61,7 @@ import org.threeten.bp.format.DateTimeParseException;
  *   <li>Float: StandardSQLTypeName.FLOAT64
  *   <li>BigDecimal: StandardSQLTypeName.NUMERIC
  *   <li>BigNumeric: StandardSQLTypeName.BIGNUMERIC
+ *   <li>JSON: StandardSQLTypeName.JSON
  * </ul>
  *
  * <p>No other types are supported through that entry point. The other types can be created by
@@ -254,6 +255,14 @@ public abstract class QueryParameterValue implements Serializable {
     return of(value, StandardSQLTypeName.STRING);
   }
 
+  /**
+   * Creates a {@code QueryParameterValue} object with a type of JSON. This is only supported in
+   * INSERT, not in query as a filter
+   */
+  public static QueryParameterValue json(String value) {
+    return of(value, StandardSQLTypeName.JSON);
+  }
+
   /** Creates a {@code QueryParameterValue} object with a type of BYTES. */
   public static QueryParameterValue bytes(byte[] value) {
     return of(value, StandardSQLTypeName.BYTES);
@@ -347,6 +356,8 @@ public abstract class QueryParameterValue implements Serializable {
       return StandardSQLTypeName.NUMERIC;
     } else if (Date.class.isAssignableFrom(type)) {
       return StandardSQLTypeName.DATE;
+    } else if (String.class.isAssignableFrom(type)) {
+      return StandardSQLTypeName.JSON;
     }
     throw new IllegalArgumentException("Unsupported object type for QueryParameter: " + type);
   }
@@ -384,6 +395,10 @@ public abstract class QueryParameterValue implements Serializable {
         break;
       case STRING:
         return value.toString();
+      case JSON:
+        if (value instanceof String) {
+          return value.toString();
+        }
       case STRUCT:
         throw new IllegalArgumentException("Cannot convert STRUCT to String value");
       case ARRAY:
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
@@ -369,8 +369,6 @@ public abstract class QueryParameterValue implements Serializable {
       return StandardSQLTypeName.JSON;
     } else if (JSONObject.class.isAssignableFrom(type)) {
       return StandardSQLTypeName.JSON;
-    } else if (Map.class.isAssignableFrom(type)) {
-      return StandardSQLTypeName.JSON;
     }
     throw new IllegalArgumentException("Unsupported object type for QueryParameter: " + type);
   }
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
@@ -29,6 +29,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.io.BaseEncoding;
+import com.google.gson.JsonObject;
 import java.io.Serializable;
 import java.math.BigDecimal;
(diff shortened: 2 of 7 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -347,6 +356,8 @@ public static QueryParameterValue struct(Map<String, QueryParameterValue> struct
       return StandardSQLTypeName.NUMERIC;
     } else if (Date.class.isAssignableFrom(type)) {
       return StandardSQLTypeName.DATE;
```
nageshs, line 358: As you mentioned, it might be useful to also allow JSONObject in addition to a string which makes constructing json easier.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -384,6 +395,10 @@ public static QueryParameterValue struct(Map<String, QueryParameterValue> struct
         break;
       case STRING:
         return value.toString();
+      case JSON:
+        if (value instanceof String) {
```
nageshs, line 399: If JSONObject support were added, I'm assuming that this too would need to be updated

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -707,6 +707,60 @@ public void testCreateTableWithRangePartitioning() {
...
+    Schema schema = Schema.of(Field.of("jsonField", StandardSQLTypeName.JSON));
+    StandardTableDefinition standardTableDefinition = StandardTableDefinition.of(schema);
+    try {
+      // Create a table with a JSON column
+      Table createdTable = bigquery.create(TableInfo.of(tableId, standardTableDefinition));
+      assertNotNull(createdTable);
+
+      // Insert data into JSON column
+      Map<String, Object> jsonRow = new HashMap<>();
+      jsonRow.put("jsonField", "{\"class\" : {\"students\" : [{\"name\" : \"Jane\"}]}}");
```
nageshs, line 723: How about inserting multiple rows each with a different name/id

e.g.
      // Insert data into JSON column
      Map<String, Object> jsonRow1 = Collections.singletonMap("jsonField", "{\"student\" : {\"name\" : \"Jane\", \"id\": 10}}");
      Map<String, Object> jsonRow2 = Collections.singletonMap("jsonField", "{\"student\" : {\"name\" : \"Joy\", \"id\": 11}}");
      Map<String, Object> jsonRow3 = Collections.singletonMap("jsonField", "{\"student\" : {\"name\" : \"Alice\", \"id\": 12}}");
      Map<String, Object> jsonRow4 = Collections.singletonMap("jsonField", "{\"student\" : {\"name\" : \"Bijoy\", \"id\": 13}}");

      InsertAllRequest request  = InsertAllRequest.newBuilder(tableId).addRow(jsonRow1).addRow(jsonRow2).addRow(jsonRow3).addRow(jsonRow4).build();

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -707,6 +707,60 @@ public void testCreateTableWithRangePartitioning() {
...
+      InsertAllRequest request = InsertAllRequest.newBuilder(tableId).addRow(jsonRow).build();
+      InsertAllResponse response = bigquery.insertAll(request);
+      assertFalse(response.hasErrors());
+      assertEquals(0, response.getInsertErrors().size());
+
+      // Query the JSON column with string positional query parameter
+      String sql =
+          "SELECT JSON_VALUE(jsonField, \"$.class.students[0].name\") FROM "
```
nageshs, line 731: The projection could then change to SELECT jsonField.class.student.id where JSON_VALUE(jsonField, \"$.class.student.name\") = ? "

id should be 10.
It may also be great to use the INSERT syntax to add the columns above with JSONObjects.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -254,6 +255,14 @@ public static QueryParameterValue string(String value) {
     return of(value, StandardSQLTypeName.STRING);
   }
 
+  /**
+   * Creates a {@code QueryParameterValue} object with a type of JSON. This is only supported in
```
nageshs, line 259: Please add: `Currently` ... this is only supported in INSERT

nageshs (commented):
Looks great!

tswast (commented):
LGTM as a first pass, but agreed with @nageshs that supporting the built-in `javax.json.JsonObject` type (automatically converting to string for the user) in addition to String would be desirable.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
tswast, line 399: I wonder if we could also support `Map<String, object>` too? That's likely what we'll be supporting in Python when we implement this.

stephaniewang526: Yes, updating this to parse JSON string also.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
stephaniewang526: Updating test case to insert all these rows (it passed).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
stephaniewang526: Done.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
nageshs: Although  Map<String, Object> is something that would be naturally used in Python, IMO,  JSONObject is more idiomatic in Java. So I'd vote to just keep JSONObject and String.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -347,6 +365,12 @@ public static QueryParameterValue struct(Map<String, QueryParameterValue> struct
       return StandardSQLTypeName.NUMERIC;
     } else if (Date.class.isAssignableFrom(type)) {
       return StandardSQLTypeName.DATE;
+    } else if (String.class.isAssignableFrom(type)) {
+      return StandardSQLTypeName.JSON;
+    } else if (JSONObject.class.isAssignableFrom(type)) {
+      return StandardSQLTypeName.JSON;
+    } else if (Map.class.isAssignableFrom(type)) {
```
nageshs, line 372: remove Map?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
stephaniewang526: Noted, thanks.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
stephaniewang526: Done.

nageshs (approved):
Thank you! Looks great to me.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
stephaniewang526: Thanks for catching it! Yes, removed.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -707,6 +708,88 @@ public void testCreateTableWithRangePartitioning() {
...
+      QueryParameterValue jsonParameter1 = QueryParameterValue.json(jsonObject);
+      QueryJobConfiguration dmlQueryJobConfiguration1 =
+          QueryJobConfiguration.newBuilder(dml)
+              .setDefaultDataset(DatasetId.of(DATASET))
+              .setUseLegacySql(false)
+              .addPositionalParameter(jsonParameter1)
+              .build();
+      bigquery.query(dmlQueryJobConfiguration1);
+      Page<FieldValueList> rows1 = bigquery.listTableData(tableId);
+      assertEquals(6, Iterables.size(rows1.getValues()));
```
tswast, line 787: Might be nice to verify that the value we get back for the `jsonField` column has the expected value.

Though hopefully the BigQuery backend at least verifies that we have inserted correct JSON syntax?

That might be an interesting test: verify that it fails if we try to send a string value that is not valid JSON.

google-cloud-bigquery/pom.xml
```diff
@@ -81,6 +81,10 @@
       <groupId>org.threeten</groupId>
       <artifactId>threetenbp</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
```
tswast, line 85: We should check with our legal team if we can use this package as a dependency.

There is a line in their license that makes this not the standard MIT open source license.

https://github.com/stleary/JSON-java/blob/f1b0210b8acdbc5f5316bc5507e2f3630506a9a3/LICENSE#L14

https://github.com/googleapis/google-http-java-client recommends Gson, and it should already be in our transitive dependencies.

tswast (changes requested):
Structure LGTM, but I have some worries about the new dependency.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
stephaniewang526: Added assert to verify JSON content.

Added another test to verify that malformed JSON is rejected.

google-cloud-bigquery/pom.xml
stephaniewang526: Thanks, removing this dependency and using `gson` instead.

tswast (approved):

PR #1865, opened 2022-02-16
docs(sample): Added AuthorizeDataset Sample

prash-mi:
Added Sample for AuthorizeDataset


samples/snippets/src/main/java/com/example/bigquery/AuthorizeDataset.java
```diff
@@ -0,0 +1,95 @@
+/*
```
tswast, line 1: Samples require integration tests to ensure they work. See: https://github.com/googleapis/java-bigquery/tree/main/samples/snippets/src/test/java/com/example/bigquery

tswast (changes requested):

stephaniewang526 (changes requested):
1) Let's put "create a dataset" prep for `userDataset` into the IT's `setUp` method. We can use `BIGQUERY_DATASET_NAME` env var as the `sourceDataset` in our sample (instead of creating another `sourceDataset`).
2) We can pass `sourceDataset` and `userDataset` into the sample method so it's clear that we are trying to authorize userDataset with sourceDataset ACL.
3) We need to delete the created `userDataset` in `cleanUp` method of the IT.

The sample body should just show only the steps to create `DatasetAclEntity` and calling dataset.update API with the new Acl list.

samples/snippets/src/main/java/com/example/bigquery/AuthorizeDataset.java
prash-mi: sure, added AuthorizeDatasetIT

prash-mi:
> 1. Let's put "create a dataset" prep for `userDataset` into the IT's `setUp` method. We can use `BIGQUERY_DATASET_NAME` env var as the `sourceDataset` in our sample (instead of creating another `sourceDataset`).
> 2. We can pass `sourceDataset` and `userDataset` into the sample method so it's clear that we are trying to authorize userDataset with sourceDataset ACL.
> 3. We need to delete the created `userDataset` in `cleanUp` method of the IT.
> 
> The sample body should just show only the steps to create `DatasetAclEntity` and calling dataset.update API with the new Acl list.

Updated the sample. Now passing the `datasetId` to the sample method and calling the update API

stephaniewang526:
Thanks! 

But this is now blocked by a backend issue: https://b.corp.google.com/issues/211633697#comment39

stephaniewang526 (approved):

prash-mi:
Closing. I will raise another PR with the required changes

PR #1879, opened 2022-03-02
Fix: adjusting retry logic to avoid retrying successful job creation

franklinWhaite:
https://github.com/googleapis/java-bigquery/pull/1744 introduced retrying 200 responses, but the regex that we are using is too broad: [".*exceed.*rate.limit."] and in some cases it may catch a valid response.

The issue is that the current logic is scanning the whole response. In this PR the retry logic is modified to specifically check the error message from the response. For this purpose parsing logic was developed to extract error messages from responses. This logic is specifically designed for the [jobs.insert method response](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert#response-body) (only case observed so far where a response with status code 200 might also return an error message).

### DIFF
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -133,7 +133,6 @@
     </dependency>
   </dependencies>
 
-
   <build>
     <plugins>
       <plugin>
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -126,7 +126,13 @@
       <artifactId>proto-google-cloud-datacatalog-v1</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/pom.xml
```diff
@@ -126,6 +126,11 @@
       <artifactId>proto-google-cloud-datacatalog-v1</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
stephaniewang526, line 129: Can we use `gson` (already included) instead? `org.json` is being banned by the Apache foundation.

stephaniewang526 (changes requested):

google-cloud-bigquery/pom.xml
franklinWhaite: Great idea! Did just that

stephaniewang526 (approved):

PR #2010, opened 2022-04-27
docs(sample): remove unused dependency and add setup instructions

mpeddada1:
Adding setup instructions that were previously missing. Removing the native-image-support module which is no longer needed.


Neenu1995 (approved):

mpeddada1:
Thank you!

Issue #2007, opened 2022-04-22
Investigate GraalVM IT failure with Apache Arrow


stephaniewang526:
We are observing GraalVM IT failure on [this PR](https://github.com/googleapis/java-bigquery/pull/1374).

Some initial investigation was done: https://github.com/googleapis/java-bigquery/pull/1374#issuecomment-1105342700

[Full stack trace](https://source.cloud.google.com/results/invocations/24a2cba4-9627-48c5-8dd6-cb880dfb0797/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-bigquery%2Fpresubmit%2Fgraalvm-native/log)

cc @suztomo

stephaniewang526:
Re-opening as the problem persists.

stephaniewang526:
Note: cl/447464498 was also done as part of this.

PR #2053, opened 2022-05-18
fix: add more native image configurations for Arrow tests and enable native image tests

mpeddada1:
Fixes https://github.com/googleapis/java-bigquery/issues/2007

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -87,7 +87,7 @@ graalvm)
     ;;
 graalvm17)
     # Run Unit and Integration Tests with Native Image. Skip Arrow tests until https://github.com/googleapis/java-bigquery/issues/2060 is fixed.
-    mvn -B ${INTEGRATION_TEST_ARGS} -ntp -Pnative -Penable-integration-tests test "-Dtest=!com.google.cloud.bigquery.it.ITBigQueryTest#testBQResultSetPaginationSlowQuery+testReadAPIConnectionMultiClose+testReadAPIIterationAndOrder, !com.google.cloud.bigquery.it.ITNightlyBigQueryTest#testIterateAndOrder+testMultipleRuns+testIterateAndOrderDefaultConnSettings"
+    mvn -B ${INTEGRATION_TEST_ARGS} -ntp -Pnative -Penable-integration-tests test "-Dtest=!com.google.cloud.bigquery.it.ITBigQueryTest#testBQResultSetPaginationSlowQuery+testReadAPIConnectionMultiClose+testReadAPIIterationAndOrder, !com.google.cloud.bigquery.it.ITNightlyBigQueryTest#testIterateAndOrder+testMultipleRuns+testIterateAndOrderDefaultConnSettings, IT*, *ClientTest"
     RETURN_CODE=$?
     ;;
 samples)
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -85,6 +85,11 @@ graalvm)
     mvn -B ${INTEGRATION_TEST_ARGS} -ntp -Pnative -Penable-integration-tests test
     RETURN_CODE=$?
     ;;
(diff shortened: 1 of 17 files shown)


mpeddada1:
Local testing succeeded so removing the "do not merge" label.

.kokoro/build.sh
```diff
@@ -80,11 +80,11 @@ nightly-it)
...
-#graalvm)
-#    # Run Unit and Integration Tests with Native Image
-#    mvn -B ${INTEGRATION_TEST_ARGS} -ntp -Pnative -Penable-integration-tests test
-#    RETURN_CODE=$?
-#    ;;
+graalvm)
+    # Run Unit and Integration Tests with Native Image
+    mvn -B ${INTEGRATION_TEST_ARGS} -ntp -Pnative -Penable-integration-tests test
+    RETURN_CODE=$?
+    ;;
```
stephaniewang526, line 87: could you also remove the `build.sh` exclusion from `owlbot.py`?

google-cloud-bigquery/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-bigquery/reflect-config.json
```diff
@@ -3,8 +3,67 @@
     "name":"io.netty.buffer.AbstractByteBufAllocator",
     "queryAllDeclaredMethods":true
   },
+  {
+    "name":"io.netty.buffer.PooledByteBufAllocator",
```
stephaniewang526, line 7: Do you intend to manually maintain this config file in the future? How much ongoing maintenance work is needed? Would we need to make these changes if we made code changes related to Arrow or add new Arrow dependencies? cc @prash-mi

stephaniewang526 (changes requested):

google-cloud-bigquery/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-bigquery/reflect-config.json
mpeddada1: That's a good question. Given that we have a bit more background on GraalVM at the moment, we plan to maintain these configurations with client library owners being involved in the review process. However, we're currently working on making our onboarding documentation more robust as we learn more about the technology so that it is easier for client library owners to contribute and get unblocked sooner if needed in the future . Will keep you posted on the plans for that.  

> Would we need to make these changes if we made code changes related to Arrow or add new Arrow dependencies?

Most of these changes involved configurations for the netty classes that Apache Arrow brings in. These currently exist in `gax-grpc` but have been added here explicitly to avoid adding a dependency on gax-grpc. Since bigquery and most libraries bring in gax, we might want to consider moving them to `gax` or even closer to netty which could help make this file a little shorter. As far as the Arrow dependencies are concerned, I don't have a solid answer for this but there is a possibility that we'll need to add new configurations if new deps are added. Luckily, we have our Kokoro GraalVM presubmit checks to catch them if required.

.kokoro/build.sh
mpeddada1: good catch, done!

owlbot.py
```diff
@@ -22,5 +22,5 @@
     '.kokoro/dependencies.sh',
     'codecov.yaml',
     'renovate.json',
-    '.kokoro/build.sh',
+    '.kokoro/build.sh'
```
stephaniewang526, line 25: This should be removed, right?

mpeddada1: I had to bring this back in to exclude the arrow tests in Java 17 until https://github.com/googleapis/java-bigquery/issues/2060 is resolved.

mpeddada1:
All checks including Kokoro JDK 17 native image are now passing.

stephaniewang526 (approved):

PR #2189, opened 2022-07-30
samples / fixing authorize dataset logic by updating source dataset acl 

nikunjbhartia:
Fixes#2188

Fixed the logic for authorizing a bigquery user dataset but updating the source dataset acl with that of the user dataset information.

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/AuthorizeDataset.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/AuthorizeDataset.java
@@ -58,12 +58,12 @@ public class AuthorizeDataset {
           new Acl.DatasetAclEntity(userDatasetId, targetTypes);
       sourceDatasetAcl.add(Acl.of(userDatasetAclEntity));
 
-      // update the user dataset with source dataset's ACL
-      Dataset updatedUserDataset =
-          userDataset.toBuilder().setAcl(sourceDatasetAcl).build().update();
+      // update the source dataset with user dataset's ACL
+      Dataset updatedSourceDataset =
+          sourceDataset.toBuilder().setAcl(sourceDatasetAcl).build().update();
 
       System.out.printf(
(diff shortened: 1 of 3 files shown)


samples/snippets/src/main/java/com/example/bigquery/AuthorizeDataset.java
```diff
@@ -58,12 +58,12 @@ public static void authorizeDataset(DatasetId sourceDatasetId, DatasetId userDat
           new Acl.DatasetAclEntity(userDatasetId, targetTypes);
       sourceDatasetAcl.add(Acl.of(userDatasetAclEntity));
 
-      // update the user dataset with source dataset's ACL
-      Dataset updatedUserDataset =
-          userDataset.toBuilder().setAcl(sourceDatasetAcl).build().update();
+      // update the source dataset with user dataset's ACL
```
averikitsch, line 61: The method comment says: "This method will update userDataset's ACL with sourceDataset's ACL". Does this need to be updated as well or is this the right change?

nikunjbhartia: Hi @averikitsch Thanks for pointing out ! 
I have updated the method. 
And yes, in this case, source dataset ACL should be updated (a similar case for authorizing a view can be found here https://github.com/googleapis/java-bigquery/blob/HEAD/samples/snippets/src/main/java/com/example/bigquery/AuthorizedViewTutorial.java#L99 )

averikitsch (approved):

nikunjbhartia:
Updated the failing test case for authorized Dataset.

nikunjbhartia:
Hi, seems this is pending review from one of the owners. Is there something I am missing ?

averikitsch:
@googleapis/api-bigquery and @googleapis/yoshi-java PTAL

PR #2222, opened 2022-08-12
fix: table-not-found issue with executeSelect while running long queries 

prash-mi:
Internal Bug's Ref: [b/241134681](url) . 

`table-not-found` issue has been observed with `executeSelect` while running long queries. This issue comes while initialising storage read session when the query job is not complete. 

This is a short term fix where we are polling the job's status using `jobs.getQueryResults` and the session with read API is initialised when the job is complete, thus avoiding `table-not-found`.

Capturing this FR for the long term fix/re-design: https://github.com/googleapis/java-bigquery/issues/2240 

Ref: go/executeselect-re-design

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
@@ -95,14 +95,6 @@ class ConnectionImpl implements Connection {
   private BlockingQueue<BigQueryResultImpl.Row>
       bufferRow; // initialized lazily iff we end up using Read API
 
-  // retry config to retry on table not found errors. Ref: b/241134681
-  private static final BigQueryRetryConfig TABLE_NOT_FOUND_RETRY_CONFIG =
-      BigQueryRetryConfig.newBuilder()
-          .retryOnMessage("Not found")
-          .retryOnMessage("NOT_FOUND")
-          .retryOnRegEx(".*not.*found.*table.*")
-          .build();
-
   // retry setting to retry on table not found errors. Settings uses a max timeout of 20 mins which
   // could be useful for really long running jobs. Ref: b/241134681
   private static RetrySettings getTableNotFoundRetrySettings() {
@@ -865,16 +857,7 @@ class ConnectionImpl implements Connection {
               .setMaxStreamCount(1) // Currently just one stream is allowed
           // DO a regex check using order by and use multiple streams
           ;
-      // Using exponential-back-off to create read session to avoid table-not-found error. Ref:
-      // b/241134681 . This approach is a short term approach and should later be replaced with a
-      // job status poll based solution to make it more deterministic.
-      ReadSession readSession =
-          BigQueryRetryHelper.runWithRetries(
-              () -> bqReadClient.createReadSession(builder.build()),
-              getTableNotFoundRetrySettings(),
-              BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
-              bigQueryOptions.getClock(),
-              TABLE_NOT_FOUND_RETRY_CONFIG);
+      ReadSession readSession = bqReadClient.createReadSession(builder.build());
 
       bufferRow = new LinkedBlockingDeque<>(getBufferSize());
       Map<String, Integer> arrowNameToIndex = new HashMap<>();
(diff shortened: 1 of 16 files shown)


Neenu1995 (approved):

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
```diff
@@ -995,33 +995,63 @@ GetQueryResultsResponse getQueryResultsFirstPage(JobId jobId) {
...
-        // with the case where there there is a HTTP error
-        throw new BigQueryException(bigQueryErrors);
+    // Implementing logic to poll the Job's status using getQueryResults as
+    // we do not get rows, rows count and schema unless the job is complete
+    // Ref: b/241134681
+    // This logic will wait for approx (poolingIntervalMs + 10 seconds which is the default timeout
+    // for getQueryResults) per iteration of the loop
+    long startTimeMs = System.currentTimeMillis();
+    long totalTimeOutMs = 18 * 60 * 60 * 1000; // 18 hours, which is the max timeout for the job
+    long poolingIntervalMs = 60000; // 1 min
```
shollyman, line 1006: nit: s/pooling/polling/

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
```diff
@@ -995,33 +995,63 @@ GetQueryResultsResponse getQueryResultsFirstPage(JobId jobId) {
...
-        // Throwing BigQueryException since there may be no JobId and we want to stay consistent
-        // with the case where there there is a HTTP error
-        throw new BigQueryException(bigQueryErrors);
+    // Implementing logic to poll the Job's status using getQueryResults as
+    // we do not get rows, rows count and schema unless the job is complete
+    // Ref: b/241134681
+    // This logic will wait for approx (poolingIntervalMs + 10 seconds which is the default timeout
+    // for getQueryResults) per iteration of the loop
+    long startTimeMs = System.currentTimeMillis();
+    long totalTimeOutMs = 18 * 60 * 60 * 1000; // 18 hours, which is the max timeout for the job
```
shollyman, line 1005: I think it's reasonable to leave the totalTimeoutMs and assorted logic out of this PR.  You may in the future want to add a user configurable timeout, but per-rpc timeouts are sufficient.

BQ guarantees that jobs make forward progress (a job won't get stuck in pending forever).  This interval is so long you may as well just trust the system.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
```diff
@@ -995,33 +995,63 @@ GetQueryResultsResponse getQueryResultsFirstPage(JobId jobId) {
...
+      } catch (BigQueryRetryHelper.BigQueryRetryHelperException e) {
+        throw BigQueryException.translateAndThrow(e);
+      }
+
+      if (results.getJobComplete()) {
+        break; // The job is complete
+
+      } else { // wait for the defined poolingIntervalMs and the loop will retry
+        try {
+          Thread.sleep(poolingIntervalMs);
```
shollyman, line 1042: The behavior here seems wrong.  getQueryResults will poll from the server side for up to 10 seconds, and then you sleep for 60 in the client?  Effective latency for an 11 sec query becomes more than a minute.

You should not wait on the client side at all, just re-issue a subsequent getQueryResults call.  Allow the poll and wait to happen within BQ.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
prash-mi: Thanks for point it out, I have update it

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
prash-mi: sure. I have updated the logic and have added a `timeoutMs` param at the RPC layer which is currently hardcoded to 60 seconds. QQ, in the very corner case when the job runs for 18 hours and is still not-complete then will the backend throw and error? Otherwise it will be an infinite loop as `jobComplete` will never be `true`. 
Also, I have captured task to make `timeoutMs` user configurable here: https://github.com/googleapis/java-bigquery/issues/2240 (point#5)

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
prash-mi: Done, updated to logic to use RCP based poll and wait

shollyman (approved):
Thanks for getting all this together!

PR #2256, opened 2022-08-29
chore: add PB methods for preserveAsciiCharaters

Neenu1995:


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/CsvOptions.java
```diff
@@ -192,6 +205,14 @@ public Long getSkipLeadingRows() {
     return skipLeadingRows;
   }
 
+  /**
+   * Returns whether BigQuery should allow ascii control characters in a CSV file. By default ascii
```
shollyman, line 209: nittiest of nits:  ASCII vs ascii for prose.

shollyman (approved):

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/CsvOptions.java
Neenu1995: Done

PR #2294, opened 2022-09-27
feat: Add executeSelectAsync and Refactor

### LINKED ISSUE #2240: Long term fix & redesign of executeSelect
b/241134681 's short term fix https://github.com/googleapis/java-bigquery/pull/2222 uses polling to fix the table_not_found issue. 
As a long term cleaner solution & redesign we can explore:

1. Providing an API Future based async implementation which can offer more control to the users (in terms of cancelling the operation and checking the status)
2. Removing patch of logic related with queryDryRun which wont be necessary after this fix
3. See if we can optimise the fast and slow query paths. The current flow is a bit convoluted . Refactor the overloaded `executeSelect` methods to avoid code duplicity.
4. A similar error can probably occur in the fast query path if we get an incomplete job id from the backend and Read API is initialised on the basis of the response from `useReadAPI` 
5. Exploring a JDBCish overloaded method [if we redesign executeSelect to return APIFuture, then we may want to have a synchronous counterpart like executeQuery as well to comply with JDBC standards]
6. Add user configurable timeout for `getQueryResults` RPC (it can be in the range 1 - 200 secs . Ref https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults)

prash-mi:
Add executeSelectAsync and Refactor

This fix addresses these tasked mentioned @ https://github.com/googleapis/java-bigquery/issues/2240

- Providing an API Future based async implementation which can offer more control to the users (in terms of cancelling the operation and checking the status)
- Removing patch of logic related with queryDryRun which wont be necessary after this fix
- See if we can optimise the fast and slow query paths. The current flow is a bit convoluted . Refactor the overloaded executeSelect methods to avoid code duplicity.
- A similar error can probably occur in the fast query path if we get an incomplete job id from the backend and Read API is initialised on the basis of the response from useReadAPI
- Exploring a JDBCish overloaded method [if we redesign executeSelect to return APIFuture, then we may want to have a synchronous counterpart like executeQuery as well to comply with JDBC standards]

Fixes #2240☕️


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
```diff
@@ -252,29 +246,136 @@ public BigQueryResult executeSelect(
...
+      String sql, List<Parameter> parameters, Map<String, String>... labels) {
+    ExecutorService execService =
+        Executors.newFixedThreadPool(
+            2); // two fixed threads. One for the async operation and the other for processing the
+    // callback
+    ListeningExecutorService lExecService = MoreExecutors.listeningDecorator(execService);
+    ListenableFuture<ExecuteSelectResponse> executeSelectFuture =
+        lExecService.submit(
+            () -> {
+              if (parameters == null) {
```
shollyman, line 303: is there value in doing this check?  does executeSelect not properly deal with null params?

shollyman (approved):
I love to see code go away. :)

Added one comment about the future construction, tests look good re: testing various failure scenarios.  I'd also like to make sure Neenu gets to review here, so please give them a chance to do so before submit.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
prash-mi: Thanks for pointing this our @shollyman , `null` check isn't required here as the underlying method deals with it. Replaced it with a call to `executeSelect(  sql,  parameters,  labels)`. Also added exception handling to pass back any exception which might have occurred.

prash-mi:
@Neenu1995 - Do you want to take a look at this when you get a chance? please let me know any comments/feedback.

Neenu1995 (approved):

PR #2349, opened 2022-10-12
feat: add fast query path support when empty jobId object is passed

Neenu1995:
Currently, if a jobID object is passed to the query method, it goes through the slow query path always. This PR allows user to set the location and project ID of the job being created with the JobID object AND take the fast query path if all the conditions for fast query path are met.

Note: If the user specifies the job id string in the job id object being supplied to the query method, the workflow will default to the slow query path and use the job id specified.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -1389,7 +1389,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
       // the job created by the query method will be in that location, even if the table to be
       // queried is in a different location. This may
       // cause the query to fail with "BigQueryException: Not found"
-      if(jobId.getLocation() != null) {
+      if (jobId.getLocation() != null) {
         content.setLocation(jobId.getLocation());
       }
       return queryRpc(projectId, content, options);
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -1384,6 +1384,12 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
     QueryRequestInfo requestInfo = new QueryRequestInfo(configuration);
     if (requestInfo.isFastQuerySupported(jobId)) {
       String projectId = getOptions().getProjectId();
+      // Be careful when setting the projectID in JobId, if a projectID is specified in the JobId,
+      // the job created by the query method will use that project. This may
+      // cause the query to fail with "BigQueryException: Not found"
+      if(jobId.getProject() != null){
+        projectId = jobId.getProject();
+      }
       QueryRequest content = requestInfo.toPb();
       // Be careful when setting the location in JobId, if a location is specified in the JobId,
       // the job created by the query method will be in that location, even if the table to be
@@ -1392,6 +1398,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
       if (jobId.getLocation() != null) {
         content.setLocation(jobId.getLocation());
       }
+
       return queryRpc(projectId, content, options);
     }
     return create(JobInfo.of(jobId, configuration), options).getQueryResults();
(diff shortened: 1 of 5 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
```diff
@@ -70,7 +70,9 @@ boolean isFastQuerySupported() {
         && config.getTableDefinitions() == null
         && config.getTimePartitioning() == null
         && config.getUserDefinedFunctions() == null
-        && config.getWriteDisposition() == null;
+        && config.getWriteDisposition() == null
+        && jobId.getJob() == null
```
shollyman, line 74: probably want a null check on jobId before checking the fields.

Neenu1995: https://github.com/googleapis/java-bigquery/blob/main/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobId.java#L48

GetJob return's the job's id

Neenu1995: Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1379,6 +1379,14 @@ public com.google.api.services.bigquery.model.QueryResponse call() {
   public TableResult query(QueryJobConfiguration configuration, JobId jobId, JobOption... options)
       throws InterruptedException, JobException {
     Job.checkNotDryRun(configuration, "query");
+    // If all parameters passed in configuration are supported by the query() method on the backend,
+    // put on fast path
+    QueryRequestInfo requestInfo = new QueryRequestInfo(configuration);
+    if (requestInfo.isFastQuerySupported(jobId)) {
```
prash-mi, line 1385: @Neenu1995 If other attributes in the `config` are set, like Clustering, TimePartitioning etc then wouldn't it take the slow path leading to Already-Exists error. Do we think if doing a `jobs.get` call and avoiding the `create` call when the job already exists could be a good idea here?

Neenu1995: I realize we can't avoid the slow path always. When the conditions are met for the fast path, then it should take fast path. 
But here, in the query method which passes job id, the only scenario where it can take the fast path is, if the location and jobid fields are not set. In such cases, the job id object is entirely disregarded. I am not sure if that is the right way to do this. What do you think?

prash-mi: Logically it is making sense to me as we are using fast query path when the jobId and location isn't passed. Couple of quick questions:

1. Is this fix supposed to be addressing the `Already-exists` issue? I am sorry if I am getting confused with that here. :)
2. Are we already covering this new flow with an Integration Test?

prash-mi (approved):
LGTM! For the given issue it will take the fast query path and it should avoid the `Already Exists` error which we saw in the slow query path. However we may want to later analyse(and maybe capture a TODO bug) if the slow query path could still give us such an error (when the operation errors out and is retried, while having any one of the given `configs` set to not `null`) . I think we should be good to merge after Seth's comments are addressed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1379,6 +1379,21 @@ public com.google.api.services.bigquery.model.QueryResponse call() {
...
+      String projectId = getOptions().getProjectId();
```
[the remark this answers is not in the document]
Neenu1995: Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequestInfo.java
```diff
@@ -57,7 +57,16 @@ final class QueryRequestInfo {
     this.useQueryCache = config.useQueryCache();
   }
 
-  boolean isFastQuerySupported() {
+  boolean isFastQuerySupported(JobId jobId) {
+    boolean isJobIdFastQuerySupported;
+    // Fast query path is not possible if job is specified in the JobID object
+    // Respect Job field value in JobId specified by user.
+    // Specifying it will force the query to take the slower path.
+    if (jobId != null) {
```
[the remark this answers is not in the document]
Neenu1995: done

PR #2379, opened 2022-10-28
deps: Remove version declaration for gson

blakeli0:


Neenu1995 (approved):

pom.xml
```diff
@@ -131,7 +131,6 @@
       <dependency>
```
Neenu1995, line 131: Since gson is dependency managed in shared-deps, removing the whole dependency from here should be fine.

blakeli0: Removed

PR #2494, opened 2023-01-26
chore: fix dataset deletion in test

Neenu1995:


alvarowolfx (approved):
Sorry for not catching that on the previous PR. Looks good to me.

PR #2509, opened 2023-02-03
fix: Move ratio calculation for whether to use read API to avoid NPE with setUseReadAPI(false)

### LINKED ISSUE #2508: BigQuery Java Client throws NPE when a query takes more than 10 seconds with useReadAPI(false)

#### Environment details

1. Specify the API at the beginning of the title. For example, "BigQuery: ...").
   General, Core, and Other are also allowed as types: Bigquery Java Client 
2. OS type and version:
3. Java version: 17 adoptium 
4.  version(s): 2.19.2 and 2.20.0 

#### Steps to reproduce

  1. set setUseReadAPI(false) on ConnectionSettings before building a connection
  2. run a query that takes more than 10 seconds using the `com.google.cloud.bigquery.Connection` query APIs. 

#### Sample Code

Standard usage with query that takes longer than 10 seconds (falls out of the fast query path). 
```
ConnectionSettings connectionSettings =
ConnectionSettings.newBuilder()
       .setUseReadAPI(false)
       .setRequestTimeout(10L)
       .setMaxResults(100L)
       .setUseQueryCache(true)
       .build();
Connection connection = bigquery.createConnection(connectionSettings);
String selectQuery = "<query that takes a little while and moves past the initial jobs.query timeout";
ListenableFuture<ExecuteSelectResponse> executeSelectFuture = connection.executeSelectAsync(selectQuery, ...);

#### Stack trace
```
java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because "totalRows" is null
	at com.google.cloud.bigquery.ConnectionImpl.useReadAPI(ConnectionImpl.java:1220)
	at com.google.cloud.bigquery.ConnectionImpl.getSubsequentQueryResultsWithJob(ConnectionImpl.java:844)
	at com.google.cloud.bigquery.ConnectionImpl.queryRpc(ConnectionImpl.java:486)
	at com.google.cloud.bigquery.ConnectionImpl.getExecuteSelectResponse(ConnectionImpl.java:236)
	at com.google.cloud.bigquery.ConnectionImpl.executeSelect(ConnectionImpl.java:218)
	at com.google.cloud.bigquery.ConnectionImpl.lambda$getExecuteSelectFuture$0(ConnectionImpl.java:305)
	at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:131)
	at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:74)
	at com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:82)
	... 3 more
```

#### Any additional information below

Looks like the exception is being thrown here: https://github.com/googleapis/java-bigquery/blob/v2.21.0/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java#L1220

Moving that check inside the if statement below it (which I think should be safe because of the check at the beginning of the method for null AND readAPI = true) should fix it. 


Following these steps guarantees the quickest resolution possible.

Thanks!

### PR DESCRIPTION
**Pull request by jonathanswenson**

Fixes #2508 ☕️

### CODE REVIEW

**Comment by Neenu1995** [comment]
@prash-mi PTAL. Thank you .

**Review by prash-mi — CHANGES_REQUESTED** [review]
Thanks for raising the PR. 
NPE shouldn't ideally come as we had made `getQueryResultFirstPage(...)` a blocking call(Ref: https://github.com/googleapis/java-bigquery/blob/main/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java#L243), such that it should not return unless the `job` is complete, which implies that we must be getting `totalRows` and `pageRows` back. 
Still we can use a `null` check IMO
Do we want to replace 
```
if ((totalRows == null || pageRows == null)
        && Boolean.TRUE.equals(
            connectionSettings
                .getUseReadAPI())) { // totalRows and pageRows are returned null when the job is not
      // complete
      return true;
    }

```
with

```
if (totalRows == null || pageRows == null) { 
      return false;
    }
```
as, we can't check the ration if any one of these attributes are `null` . WDYT?

**Comment by jonathanswenson** [comment]
@prash-mi That sounds reasonable to me, I have made that additional change. Let me know what you think.

**Review by prash-mi — APPROVED** [review]
thanks for making the changes, LGTM

**Comment by obada-ab** [comment]
@jonathanswenson thanks for making the changes, it's better to add unit tests to the changes to ensure that the NPE would be avoided

**Comment by obada-ab** [comment]
@prash-mi there's a behaviour change as a side effect of this change that I'm worried about. Specifically, when either `totalRows` or `pageRows` are `null` (due to the job taking a long time) and  `setUseReadAPI(true)` is used in the connection settings.

The current behaviour would be to use the Read API, but after introducing this fix such queries would always use `tabledata.list`.

If you believe that this change makes sense, then we can unit test and merge it. Otherwise maybe the following change might be reasonable:

```
    // Read API does not yet support Interval Type or QueryParameters
    if (containsIntervalType(schema) || hasQueryParameters) {
      logger.log(Level.INFO, "\n Schema has IntervalType, or QueryParameters. Disabling ReadAPI");
      return false;
    }

    if (totalRows == null || pageRows == null) {
      return connectionSettings.getUseReadAPI();
    }
    
    if (Boolean.TRUE.equals(connectionSettings.getUseReadAPI())) {
      long resultRatio = totalRows / pageRows;
      return resultRatio >= connectionSettings.getTotalToPageRowCountRatio()
          && totalRows > connectionSettings.getMinResultSize();
    } else {
      return false;
    }
```

This would avoid the NPE while maintaining the current behaviour.

**Comment by obada-ab** [comment]
Added mentioned change alongside unit testing. I don't think new integration tests are useful, but it might be worth it to run the benchmark queries to ensure long-running jobs are working fine, and that their execution time hasn't gotten worse. Will work on that.

**Comment by prash-mi** [comment]
@obada-ab I think returning `connectionSettings.getUseReadAPI()` if `totalRows`  or `pageRows`  is `null` is the right thing to do.
Thanks for making changes, LGTM!

**Comment by obada-ab** [comment]
I ran the benchmark queries locally against my test project. The third run of the queries in `main` gave these results:

```
query "SELECT * FROM `nyc-tlc.yellow.trips` LIMIT 10000": read 10000 rows, 19 cols, first byte 3.907000 sec, total 3.910000 sec
query "SELECT * FROM `nyc-tlc.yellow.trips` LIMIT 100000": read 100000 rows, 19 cols, first byte 5.695000 sec, total 25.436000 sec
query "SELECT * FROM `nyc-tlc.yellow.trips` LIMIT 1000000": read 1000000 rows, 19 cols, first byte 6.617000 sec, total 239.012000 sec
query "SELECT title FROM `bigquery-public-data.samples.wikipedia` ORDER BY title LIMIT 1000": read 1000 rows, 1 cols, first byte 0.664000 sec, total 0.664000 sec
query "SELECT title, id, timestamp, contributor_ip FROM `bigquery-public-data.samples.wikipedia` WHERE title like 'Blo%' ORDER BY id": read 193029 rows, 4 cols, first byte 7.252000 sec, total 13.599000 sec
query "SELECT * FROM `bigquery-public-data.baseball.games_post_wide` ORDER BY gameId": read 8676 rows, 145 cols, first byte 4.839000 sec, total 11.641000 sec
query "SELECT * FROM `bigquery-public-data.samples.github_nested` WHERE repository.has_downloads ORDER BY repository.created_at LIMIT 10000": read 10000 rows, 8 cols, first byte 7.865000 sec, total 12.966000 sec
query "SELECT repo_name, path FROM `bigquery-public-data.github_repos.files` WHERE path LIKE '%.java' ORDER BY id LIMIT 1000000": read 1000000 rows, 2 cols, first byte 12.123000 sec, total 110.380000 sec
```

The third run in `fix-read-query-npe` gave these results:

```
query "SELECT * FROM `nyc-tlc.yellow.trips` LIMIT 10000": read 10000 rows, 19 cols, first byte 4.089000 sec, total 4.094000 sec
query "SELECT * FROM `nyc-tlc.yellow.trips` LIMIT 100000": read 100000 rows, 19 cols, first byte 5.584000 sec, total 25.566000 sec
query "SELECT * FROM `nyc-tlc.yellow.trips` LIMIT 1000000": read 1000000 rows, 19 cols, first byte 5.847000 sec, total 238.485000 sec
query "SELECT title FROM `bigquery-public-data.samples.wikipedia` ORDER BY title LIMIT 1000": read 1000 rows, 1 cols, first byte 0.602000 sec, total 0.603000 sec
query "SELECT title, id, timestamp, contributor_ip FROM `bigquery-public-data.samples.wikipedia` WHERE title like 'Blo%' ORDER BY id": read 193029 rows, 4 cols, first byte 6.885000 sec, total 13.119000 sec
query "SELECT * FROM `bigquery-public-data.baseball.games_post_wide` ORDER BY gameId": read 8676 rows, 145 cols, first byte 4.057000 sec, total 10.648000 sec
query "SELECT * FROM `bigquery-public-data.samples.github_nested` WHERE repository.has_downloads ORDER BY repository.created_at LIMIT 10000": read 10000 rows, 8 cols, first byte 8.350000 sec, total 13.840000 sec
query "SELECT repo_name, path FROM `bigquery-public-data.github_repos.files` WHERE path LIKE '%.java' ORDER BY id LIMIT 1000000": read 1000000 rows, 2 cols, first byte 11.803000 sec, total 107.445000 sec
```

Nothing concerning IMO, the differences seem minor and are probably just due to variance between runs.

obada-ab (approved):

PR #2553, opened 2023-03-02
feat: add support for clone

Neenu1995:


Neenu1995:
@shollyman PTAL.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -5179,4 +5181,58 @@ public void testCreateExternalTableWithReferenceFileSchemaParquet() {
...
+    CopyJobConfiguration createdConfiguration = createdJob.getConfiguration();
+    assertNotNull(createdConfiguration.getSourceTables());
+    assertNotNull(createdConfiguration.getOperationType());
+    assertNotNull(createdConfiguration.getDestinationTable());
+    Job completedJob = createdJob.waitFor();
+    assertNull(completedJob.getStatus().getError());
+
+    Table cloneTable = bigquery.getTable(DATASET, cloneTableName);
+    assertNotNull(cloneTable);
+    System.out.println(cloneTable.getDefinition().getType());
```
shollyman, line 5223: Guessing the System.out.println lines in here are stale debug statements that survived?

shollyman (approved):
Thanks for adding this!  I'm always a little hesistant with extensions to TableDefinition due to it being a java-only pattern.  I wonder if we should have a generic mechanism somewhere for populating the known common fields within the various table definition subclasses.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
Neenu1995: Yes

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableDefinition.java
```diff
@@ -85,6 +85,8 @@ public Type apply(String constant) {
 
     public static final Type SNAPSHOT = type.createAndRegister("SNAPSHOT");
 
+    public static final Type CLONE = type.createAndRegister("CLONE");
```
shollyman, line 88: Oh, here's the issue.  We likely shouldn't make a new CloneTableDefinition at all.  The type returned by the table metadata for a clone tables is still TABLE, not CLONE.  It just has the CloneTableDefinition present.

Neenu1995: Got it.. That Makes a lot more sense. By that definition shouldn't SNAPSHOT also follow a similar implementation flow? Or is that different?
Does creating a SNAPSHOT table change its definition?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
```diff
@@ -205,6 +211,7 @@ static class BuilderImpl extends Builder {
       this.labels = Labels.fromPb(tablePb.getLabels());
       this.requirePartitionFilter = tablePb.getRequirePartitionFilter();
       this.defaultCollation = tablePb.getDefaultCollation();
+      this.cloneDefinition = CloneDefinition.fromPb(tablePb.getCloneDefinition());
```
shollyman, line 214: Not sure if you want it here or elsewhere in fromPb, but don't you need to handle null cloneDefinition?

Neenu1995: Fixed

PR #2689, opened 2023-05-11
feat: return JobID with TableResult

Neenu1995:
Currently there is no easy way to log the successful JobID after the job has been completed. Prompting customers to pass their own [jobIDs for the query method](https://github.com/googleapis/java-bigquery/blob/193980319cb743e6b6c67648ddb21432e5e69ff8/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java#L1385). If the JobID is returned with the result of the query method, customers don’t have to pass the explicit JobID anymore, enabling more customers to use the fast query path.
 
Query method returns a TableResult  object. Adding a JobID field to the TableResult could convince customers to use the [query method overload](https://github.com/googleapis/java-bigquery/blob/193980319cb743e6b6c67648ddb21432e5e69ff8/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java#L1296) without passing explicit JobID.


Neenu1995:
@shollyman PTAL.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1355,7 +1355,10 @@ public com.google.api.services.bigquery.model.QueryResponse call() {
       // here, but this is left as future work.
       JobId jobId = JobId.fromPb(results.getJobReference());
       Job job = getJob(jobId, options);
-      return job.getQueryResults();
+      TableResult tableResult = job.getQueryResults();
+      // Return the JobID of the successful job
+      tableResult.setJobId(jobId);
```
shollyman, line 1360: Is the explicit setJobId even needed here?  Won't getQueryResults set it automatically, as its already got the awareness and has the special logic for results vs empty resultsets, etc?

Neenu1995: Agreed. Resolved the redundancy.

shollyman (approved):

PR #2698, opened 2023-05-15
fix: add support for repeated record query parameters

### LINKED ISSUE #2485: Bigquery: array-of-structs not working as named parameter in the java library
#### Environment details

2. OS type and version: Mac OS 
3. Java version: 17
4.  version(s):
```
bq version:
This is BigQuery CLI 2.0.83

google-cloud-bigquery library: 2.18.2
```
#### Steps to reproduce

  1. execute the bq -command line code below
  2. execute the equivalent java code, which fails with ` Cannot convert STRUCT to String value`

#### Code example

it works to supply a parameter as an array of structs:
```
bq query \
    --use_legacy_sql=false \          
    --parameter='tupels:ARRAY<STRUCT<a INT64, b INT64>>:[{"a":1, "b": 2}]' \
    'SELECT * FROM (SELECT 1 as a, 2 as b) WHERE (a, b) IN UNNEST(@tupels)'
```

works, while it does not seem to be possible to construct the same query with the java api:

```
  public static void queryWithArrayOfStructsNamedParams() {
    try {
      // Initialize client that will be used to send requests. This client only needs to be created
      // once, and can be reused for multiple requests.
      BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();

      // Create struct and add it to the list
      Map<String, QueryParameterValue> structMap = new HashMap<>();
      structMap.put("a", QueryParameterValue.int64(1));
      structMap.put("b", QueryParameterValue.int64(2));

      QueryParameterValue structQueryParam = QueryParameterValue.struct(structMap);
      List<QueryParameterValue> tupels = new ArrayList<>();
      tupels.add(structQueryParam);

      String query = "SELECT * FROM (SELECT 1 as a, 2 as b) WHERE (a, b) IN UNNEST(@tupels)";
      QueryJobConfiguration queryConfig =
          QueryJobConfiguration.newBuilder(query)
              .setUseLegacySql(false)
              .addNamedParameter("tupels", QueryParameterValue.array(tupels.toArray(), StandardSQLTypeName.STRUCT))
              .build();
      TableResult results = bigquery.query(queryConfig);
      results
          .iterateAll()
          .forEach(row -> row.forEach(val -> logger.info(val.toString())));
      System.out.println("Query with Array of struct parameter performed successfully.");
    } catch (Exception e) {
      System.out.println("Query not performed \n" + e);
    }
  }
```

#### Stack trace

the above code gives this stack trace
```
java.lang.IllegalArgumentException: Cannot convert STRUCT to String value
	at com.google.cloud.bigquery.QueryParameterValue.valueToStringOrNull(QueryParameterValue.java:440)
	at com.google.cloud.bigquery.QueryParameterValue.of(QueryParameterValue.java:216)
	at com.google.cloud.bigquery.QueryParameterValue.array(QueryParameterValue.java:352)
// ... custom code
```

#### External references such as API reference guides

- https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays
- https://cloud.google.com/bigquery/docs/parameterized-queries
- sample inspired by https://github.com/googleapis/java-bigquery/issues/2171

obada-ab:
Add support for array of struct query parameters which are used to query repeated record fields

Fixes #2485 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
@@ -37,6 +37,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import javax.annotation.Nullable;
 import org.threeten.bp.Instant;
 import org.threeten.bp.ZoneOffset;
@@ -349,7 +350,11 @@ public abstract class QueryParameterValue implements Serializable {
   public static <T> QueryParameterValue array(T[] array, StandardSQLTypeName type) {
     List<QueryParameterValue> listValues = new ArrayList<>();
     for (T obj : array) {
-      listValues.add(QueryParameterValue.of(obj, type));
+      if (type == StandardSQLTypeName.STRUCT) {
+        listValues.add((QueryParameterValue) obj);
+      } else {
+        listValues.add(QueryParameterValue.of(obj, type));
+      }
     }
     return QueryParameterValue.newBuilder()
         .setArrayValues(listValues)
@@ -522,9 +527,16 @@ public abstract class QueryParameterValue implements Serializable {
     QueryParameterType typePb = new QueryParameterType();
     typePb.setType(getType().toString());
     if (getArrayType() != null) {
-      QueryParameterType arrayTypePb = new QueryParameterType();
-      arrayTypePb.setType(getArrayType().toString());
-      typePb.setArrayType(arrayTypePb);
+      if (getArrayType() == StandardSQLTypeName.STRUCT) {
+        List<QueryParameterValue> values =
+            Objects.requireNonNull(getArrayValues(), "Array of struct cannot be empty");
+        QueryParameterType structType = values.get(0).toTypePb();
+        typePb.setArrayType(structType);
+      } else {
+        QueryParameterType arrayTypePb = new QueryParameterType();
+        arrayTypePb.setType(getArrayType().toString());
+        typePb.setArrayType(arrayTypePb);
+      }
     }
     if (getStructTypes() != null) {
       List<QueryParameterType.StructTypes> structTypes = new ArrayList<>();
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
@@ -4063,6 +4063,47 @@ public class ITBigQueryTest {
   }
 
   @Test
+  public void testRepeatedRecordNamedQueryParameters() throws InterruptedException {
+    String[] stringValues = new String[] {"test-stringField", "test-stringField2"};
+    List<QueryParameterValue> tuples = new ArrayList<>();
+    for (int i = 0; i < 2; i++) {
+      QueryParameterValue stringValue = QueryParameterValue.string(stringValues[i]);
+      Map<String, QueryParameterValue> struct = new HashMap<>();
+      struct.put("stringField", stringValue);
+      QueryParameterValue recordValue = QueryParameterValue.struct(struct);
+      tuples.add(recordValue);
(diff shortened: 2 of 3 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -522,9 +527,16 @@ QueryParameterType toTypePb() {
     QueryParameterType typePb = new QueryParameterType();
     typePb.setType(getType().toString());
     if (getArrayType() != null) {
-      QueryParameterType arrayTypePb = new QueryParameterType();
-      arrayTypePb.setType(getArrayType().toString());
-      typePb.setArrayType(arrayTypePb);
+      if (getArrayType() == StandardSQLTypeName.STRUCT) {
+        List<QueryParameterValue> values =
+            Objects.requireNonNull(getArrayValues(), "Array of struct cannot be empty");
```
Neenu1995, line 532: Where is this restriction coming from?

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -4062,6 +4062,47 @@ public void testStructNamedQueryParameters() throws InterruptedException {
     }
   }
 
+  @Test
+  public void testRepeatedRecordNamedQueryParameters() throws InterruptedException {
```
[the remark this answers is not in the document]
obada-ab: Sure, will work on it

obada-ab:
@Neenu1995 I've made some changes:

* The `toTypePb` method now only peeks at struct array values if they exist, otherwise it defaults to generating an empty struct array type
* Added two additional integration tests:
  * `testUnnestRepeatedRecordNamedQueryParameter`: tests UNNESTing a struct array query parameter
  * `testEmptyRepeatedRecordNamedQueryParameters`: tests creating an empty struct array query parameter. Creating the query parameter itself works but executing `IN UNNEST` for said parameter should result in an exception

Also there's something to note: the `toTypePb` is only used for query parameters, returned query results only go through `toValuePb` which wasn't touched in this PR.

Neenu1995:
```
String query = "SELECT string, arrayOfStruct FROM `playground-argolis.demo.allTypeTable`, "
          + "UNNEST(arrayOfStruct) as TEMP WHERE TEMP IN UNNEST(@states);";

      // states - Parameterized query type of Array of Struct
      // arrayOfStruct - Field of type Array of Struct
```

This is the exact statement we want to verify. Verify if a named parameter type of Array of struct exists in a struct field.
Please create a table and try querying from there.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -4142,6 +4142,61 @@ public void testUnnestRepeatedRecordNamedQueryParameter() throws InterruptedExce
...
+  public void testUnnestRepeatedRecordNamedQueryParameterFromDataset() throws InterruptedException {
+    List<QueryParameterValue> tuples = new ArrayList<>();
+      QueryParameterValue statusValue = QueryParameterValue.string("single");
+    QueryParameterValue addressValue = QueryParameterValue.string("123 this lane");
+    QueryParameterValue cityValue = QueryParameterValue.string("Torono");
+    QueryParameterValue stateValue = QueryParameterValue.string("ON");
+        QueryParameterValue zipValue = QueryParameterValue.string("1h2j34");
+    QueryParameterValue numberOfYearsValue = QueryParameterValue.string("3");
+
+      Map<String, QueryParameterValue> struct = new HashMap<>();
```
obada-ab: The order of fields inside structs matters when doing comparisons, and `HashMap` does not maintain insertion order. Because of these reasons, the `IN UNNEST(addresses)` query doesn't return any results.

I tried replacing `HashMap` with a `LinkedHashMap` and using `ImmutableMap.of`, both options made the `IN UNNEST(addresses)` query work as intended and return the correct row.

PR #2876, opened 2023-09-13
fix: update samples snippet to write to BYTES instead of ARRAY<BYTES>

### LINKED ISSUE #2875: Samples snippet InsertingDataTypes.java is writing byte array to `ARRAY<BYTES>` instead of `BYTES`
Samples snippet : https://github.com/googleapis/java-bigquery/blob/main/samples/snippets/src/main/java/com/example/bigquery/InsertingDataTypes.java is writing a byte array to `ARRAY<BYTES>` column in bigquery (BYTES in REPEATED mode) which gets stored as multiple byte arrays in BigQuery and becomes unserializable on reading.

Currently 

- BigQuery java client doesn't support directly writing byte array to `BYTES` datatype in bigquery, but converting that byte array to base64 string and then writing via client works smoothly
- BigQuery java client currently supports writing byte array to `ARRAY<BYTES>` which should not supported

forksumit:
Update samples snippet [InsertingDataTypes.java](https://github.com/googleapis/java-bigquery/blob/main/samples/snippets/src/main/java/com/example/bigquery/InsertingDataTypes.java)
- Change datatype for `school` column for REPEATED BYTES to BYTES
- Convert byte array to base64 string when writing into BYTES column (`school`)

Fixes #2875 ☕️

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquery/InsertingDataTypes.java
+++ b/samples/snippets/src/main/java/com/example/bigquery/InsertingDataTypes.java
@@ -30,6 +30,7 @@ import com.google.cloud.bigquery.StandardTableDefinition;
 import com.google.cloud.bigquery.TableDefinition;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableInfo;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


obada-ab (approved):

obada-ab:
LGTM, manually tested and after the change a correct base64 encoded string is being uploaded to BQ

PR #2890, opened 2023-09-19
chore: remove unused release jobs

alicejli:
We have moved the release jobs to Google3. The requirements.txt
is now unused here.

Internal bug: b/296905003

Note that I had to add entries in owlbot.py because they are templated files.


alicejli:
Closing this - they were removed in https://github.com/googleapis/java-bigquery/pull/3095

PR #2913, opened 2023-09-29
feat: Add support for Dataset property storageBillingModel

brendangreenley:
Adds support for setting and getting the BigQuery v2 API Dataset property `storageBillingModel`, to allow end users to set the billing model depending on their use case. When no billing model is provided, the existing behavior (use the default billing model) continues.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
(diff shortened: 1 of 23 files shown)


obada-ab:
Hey @brendangreenley can you add a test in ITBigQueryTest.java?

brendangreenley:
> Hey @brendangreenley can you add a test in ITBigQueryTest.java?

@obada-ab  Gave it a go, let me know if I need to adjust anything

obada-ab (approved):

PR #2923, opened 2023-10-05
feat: add support for preview features

obada-ab:
Enables preview query features, which currently only includes stateless queries (queries without jobId). These features won't always be enabled on the service side and there are additional checks and conditions.

Fixes https://github.com/googleapis/java-bigquery/issues/2949

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
@@ -95,9 +95,18 @@ public final class QueryJobConfiguration extends JobConfiguration {
     BATCH
   }
 
-  public enum JobCreationMode {
+  /** Job Creation Mode provides different options on job creation. */
+  enum JobCreationMode {
+    /** Unspecified JobCreationMode, defaults to JOB_CREATION_REQUIRED. */
     JOB_CREATION_MODE_UNSPECIFIED,
+    /** Default. Job creation is always required. */
     JOB_CREATION_REQUIRED,
+    /**
+     * Job creation is optional. Returning immediate results is prioritized. BigQuery will
+     * automatically determine if a Job needs to be created. The conditions under which BigQuery can
+     * decide to not create a Job are subject to change. If Job creation is required,
+     * JOB_CREATION_REQUIRED mode should be used, which is the default.
+     */
     JOB_CREATION_OPTIONAL,
   }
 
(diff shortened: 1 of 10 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -94,6 +95,21 @@ public enum Priority {
     BATCH
   }
 
+  /** Job Creation Mode provides different options on job creation. */
+  enum JobCreationMode {
```
shollyman, line 99: So, I think the interesting question is whether we want to expose the enum and setter at all to users.  It's only available on the jobs.query path so it feels a bit unusual to overload JobConfiguration with this.

Originally I envisioned the only control over the setting being the env variable, and the library would set it appropriately when we're on the fast query path.  After preview we'd simply always set it on the fast query path, much like how the idempotence request ID is handled.  If you truly want a job to be created, you can invoke the proper method that returns a job rather than an iterator.

The issue manifests here in some way as how the enum is documented. The default really isn't static in terms of the setter, as the client library is going to be setting values automatically so you need to potentially describe this either more ambiguously, or possibly use JOB_CREATION_MODE_UNSPECIFIED to describe the situations in which the library will set one value vs another.

I'm not sure I hold particularly strong views here, so I'd love to hear your thoughts.

obada-ab: @shollyman thanks for taking a look and bringing this up

I'm not entirely sure what you mean by "expose the enum and setter at all to users", users wouldn't be able to use the setter `setJobCreationMode` or the enum `JobCreationMode` since both are package-private. The only (relevant) external method added in this change is `setQueryPreviewEnabled` which I used for integration tests. 

I added the enum/setter/documentation to make the internal client dev process a bit better, I am not sure how/when this part of the code would be changed/removed later on though.

If you still think some parts of the code/documentation should be removed please let me know.

obada-ab:
@shollyman @PhongChuong friendly ping

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
shollyman: My apologies, I missed the package private.  That makes more sense.

shollyman (approved):

PR #2999, opened 2023-11-13
fix: update TableInsertRows.java

### LINKED ISSUE #3000: Use row id for addRows in the sample TableInsertRows.java
The current sample [TableInsertRows.java](https://github.com/googleapis/java-bigquery/blob/v2.34.2/samples/snippets/src/main/java/com/example/bigquery/TableInsertRows.java) doesn't explain how to set row ids in `InsertAllRequest.addRow`.  I think it should shows how to set row ids at Streaming Inserts.

* As [TableInsertRowsWithoutRowIds.java](https://github.com/googleapis/java-bigquery/blob/v2.34.2/samples/snippets/src/main/java/com/example/bigquery/TableInsertRowsWithoutRowIds.java) explains how-to without row ids, I believe TableInsertRows.java should explain how-to with row ids.
* Setting no row ids has an unexpected behaviors that retry is disable (b/280865468).

baeminbo:
Make the example set row id in `addRow`. If row id is missed, it disable the retry b/280865468, which I believe an unexpected behavior to users.

Fixes #3000.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ If you are using Maven without the BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies:
 
 ```Groovy
-implementation platform('com.google.cloud:libraries-bom:26.26.0')
+implementation platform('com.google.cloud:libraries-bom:26.27.0')
 
 implementation 'com.google.cloud:google-cloud-bigquery'
 ```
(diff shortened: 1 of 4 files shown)


Neenu1995:
@baeminbo Could you also fix the sample test to match the changes in code?

baeminbo:
Hi, I updated `TableInsertRowsIT`. Now, I'm waiting on the integration test result.

Neenu1995 (approved):

PR #3015, opened 2023-11-27
feat: Add support for Dataset property maxTimeTravelHours

brendangreenley:
Adds support for setting and getting the BigQuery v2 API Dataset property maxTimeTravelHours, allowing end users to set the default time travel length for a given dataset. Exposing this flag will allow smarter values to be set for staging datasets which are deleted shortly after creation, which can be costly if a user has pursued a physical billing model for their project/organization.

I used a Long since the [googleapis/google-api-java-client-services implementation](https://github.com/googleapis/google-api-java-client-services/blob/7a9657c76e15de72ba419f76cf112543ea1f53de/clients/google-api-services-bigquery/v2/2.0.0/com/google/api/services/bigquery/model/Dataset.java#L196) uses a Long.


brendangreenley:
@mrfaizal please let me know if there's anything I can do to help get this merged

PR #3047, opened 2023-12-15
feat: Use location in BigQueryOption as the default for query

### LINKED ISSUE #3041: Location not set properly in stateless query path
The location set in `BigQueryOptions` is not being respected in low-latency query path.

PhongChuong:

Fixes #3041 ☕️

### DIFF
--- a/.github/.OwlBot.lock.yaml
+++ b/.github/.OwlBot.lock.yaml
@@ -13,4 +13,5 @@
 # limitations under the License.
 docker:
   image: gcr.io/cloud-devrel-public-resources/owlbot-java:latest
-  digest: sha256:bc2bf47c7370f1b1a8a46b0c104ce7e43644ac58902c9de265fe1f253fcc2506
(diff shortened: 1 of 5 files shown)


obada-ab:
@PhongChuong I think this change can be tested in `ITBigQueryTest` by trying to query a dataset that is in a specific location, for example if the location set in `BigQueryOptions` is `US` then attempting to query an `EU` dataset should fail. Something similar is being done [here](https://github.com/googleapis/java-bigquery/blob/ff93fb8c7832ed01b59ca66f34510e343957bc83/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java#L5578). What do you think?

PhongChuong:
@obada-ab , thanks.  Added testStatelessQueriesWithLocation integration test. Please take a look.

obada-ab (approved):

PR #3087, opened 2024-01-22
Feat: Add universe domain

### LINKED ISSUE #3088: Support universe domain feature
Support setting the universe domain.

PhongChuong:
This PR adds the ability to set the universe domain and validate that the parameters are correct prior to each RPC. This is done by wrapping the existing RPC with the provided validation ServiceOption method. The PR does not change existing functionality of BigQuery and all existing unit and integration tests passes for the default universe domain. Manual testing was performed on a non-default universe domain. However, IT will require additional work (as it require additional credential setup) and will be included in a future PR.

Fixes #3088  ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -52,6 +52,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -269,7 +270,8 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
     try {
       return Dataset.fromPb(
           this,
-          runWithRetries(
+          BigQueryRetryHelper.validateAndRunWithRetries(
+              getOptions(),
               new Callable<com.google.api.services.bigquery.model.Dataset>() {
                 @Override
                 public com.google.api.services.bigquery.model.Dataset call() {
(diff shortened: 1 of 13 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1125,8 +1210,20 @@ public TableDataInsertAllResponse call() throws Exception {
                 getOptions().getClock());
       } catch (RetryHelperException e) {
         throw BigQueryException.translateAndThrow(e);
+      } catch (IllegalArgumentException | IOException e) {
+        // Invalid universe domain exceptions.
+        throw BigQueryException.translateAndThrow(e);
       }
     } else {
+      // This calls the backend without the retry wrapper so first validate the universe domain.
```
shollyman, line 1218: for the purposes of standardizing the pattern here, is it worth using validateAndRunWithRetries here, but passing in a static retry config that only allows a single invocation rather than using `getOptions().getRetrySettings()`?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1263,30 +1368,36 @@ public Page<Job> listJobs(JobListOption... options) {
...
+          Iterables.transform(
+              result.y(),
+              new Function<com.google.api.services.bigquery.model.Job, Job>() {
+                @Override
+                public Job apply(com.google.api.services.bigquery.model.Job job) {
+                  return Job.fromPb(serviceOptions.getService(), job);
+                }
+              });
+      return new PageImpl<>(new JobPageFetcher(serviceOptions, cursor, optionsMap), cursor, jobs);
+    } catch (IllegalArgumentException | IOException e) {
```
shollyman, line 1397: Do we need to catch/translate the potential RetryHelper.RetryHelperException (even though we didn't appear to previously)?

shollyman (commented):
Two things:

* testing appears to be missing is something that exercises the invalid combination of mismatched universe in credentials and service option.

* ensure that the host/endpoint functions properly when issuing rpcs, though this may make more sense out of band as a smoke test until we have a better way of doing universe integration testing.

PR #3106, opened 2024-01-30
Feat: Add queryId to TableResult

### LINKED ISSUE #3105: Add support for queryId in TableResult

PhongChuong:
This PR adds queryId to TableResult by changing its public constructors. It should be fine to update TableResult constructors to include QueryId since it is used to return results to the user and users should not directly construct TableResult objects.

Fixes #3105 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
@@ -37,6 +37,8 @@ public class TableResult implements Page<FieldValueList>, Serializable {
   private final Page<FieldValueList> pageNoSchema;
   @Nullable private JobId jobId = null;
 
+  @Nullable private final String queryId;
+
   // package-private so job id is not set outside the package.
   void setJobId(@Nullable JobId jobId) {
     this.jobId = jobId;
@@ -46,24 +48,35 @@ public class TableResult implements Page<FieldValueList>, Serializable {
     return jobId;
   }
 
+  public String getQueryId() {
+    return queryId;
+  }
+
   /**
    * If {@code schema} is non-null, {@code TableResult} adds the schema to {@code FieldValueList}s
    * when iterating through them. {@code pageNoSchema} must not be null.
    */
   @InternalApi("Exposed for testing")
-  public TableResult(Schema schema, long totalRows, Page<FieldValueList> pageNoSchema) {
+  public TableResult(
+      Schema schema, long totalRows, Page<FieldValueList> pageNoSchema, String queryId) {
     this.schema = schema;
     this.totalRows = totalRows;
     this.pageNoSchema = checkNotNull(pageNoSchema);
+    this.queryId = queryId;
   }
 
   @InternalApi("Exposed for testing")
   public TableResult(
-      Schema schema, long totalRows, Page<FieldValueList> pageNoSchema, JobId jobId) {
+      Schema schema,
+      long totalRows,
+      Page<FieldValueList> pageNoSchema,
+      JobId jobId,
+      String queryId) {
     this.schema = schema;
     this.totalRows = totalRows;
     this.pageNoSchema = checkNotNull(pageNoSchema);
     this.jobId = jobId;
+    this.queryId = queryId;
   }
 
   /** Returns the schema of the results. Null if the schema is not supplied. */
@@ -92,7 +105,7 @@ public class TableResult implements Page<FieldValueList>, Serializable {
   @Override
   public TableResult getNextPage() {
     if (pageNoSchema.hasNextPage()) {
-      return new TableResult(schema, totalRows, pageNoSchema.getNextPage());
+      return new TableResult(schema, totalRows, pageNoSchema.getNextPage(), queryId);
     }
     return null;
   }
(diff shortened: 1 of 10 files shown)


PhongChuong:
Please let me know if it's not advisable to update the constructor for TableResult.

tswast:
> Please let me know if it's not advisable to update the constructor for TableResult.

I recommend asking on a Java chat. This may be considered a breaking change.

PhongChuong:
> > Please let me know if it's not advisable to update the constructor for TableResult.
> 
> I recommend asking on a Java chat. This may be considered a breaking change.

Just verified, since the constructors are annotated with
@InternalApi("Exposed for testing")
It should be fine to change the constructor.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -6314,19 +6314,27 @@ public void testStatelessQueries() throws InterruptedException {
...
-    assertNull(executeSimpleQuery(bigQuery).getJobId());
+    tableResult = executeSimpleQuery(bigQuery);
+    assertNotNull(tableResult.getQueryId());
+    assertNull(tableResult.getJobId());
 
     // any other values won't enable optional job creation mode
     bigQuery.getOptions().setQueryPreviewEnabled("test_value");
-    assertNotNull(executeSimpleQuery(bigQuery).getJobId());
+    tableResult = executeSimpleQuery(bigQuery);
+    assertNotNull(tableResult.getQueryId());
```
Neenu1995, line 6330: How does this scenario work? Will the jobID be populated when the queryID is populated?
Isn't the jobID and queryID fields an either..or case?

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -6314,19 +6314,27 @@ public void testStatelessQueries() throws InterruptedException {
...
+    assertNotNull(tableResult.getJobId());
 
     // reset the flag
     bigQuery.getOptions().setQueryPreviewEnabled(null);
-    assertNotNull(executeSimpleQuery(bigQuery).getJobId());
+    tableResult = executeSimpleQuery(bigQuery);
+    assertNotNull(tableResult.getQueryId());
+    assertNotNull(tableResult.getJobId());
```
Neenu1995, line 6337: Same question here.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
PhongChuong: It seems like queryId is always populated. As far as I can tell, the queryId == jobId string when jobId is not null.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
PhongChuong: ditto.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
```diff
@@ -46,24 +48,35 @@ public JobId getJobId() {
...
+    return queryId;
+  }
+
   /**
    * If {@code schema} is non-null, {@code TableResult} adds the schema to {@code FieldValueList}s
    * when iterating through them. {@code pageNoSchema} must not be null.
    */
   @InternalApi("Exposed for testing")
-  public TableResult(Schema schema, long totalRows, Page<FieldValueList> pageNoSchema) {
+  public TableResult(
```
shollyman, line 60: Rather than solely extending the existing constructor, should we just provide an second signature or a builder?  It would mean you can avoid having to touch all the callsites where queryId isn't relevant for construction.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
shollyman: expectations:  
* all queries run through jobs.query statelessly should return a queryId.
* if we run through jobs.query and end up with a stateful (job-based) execution then I currently expect both to be populated.
* If we run via jobs.insert I only expect jobID because queryId isn't part of the job message (or its submessages like statistics).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
PhongChuong: As discussed offline. I think creating a builder would be the preferred option. I'll follow up with another PR to switch to using a builder.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
PhongChuong: I added an additional test: testTableResultJobIdAndQueryId which runs through the scenario @shollyman described. There is a slight difference in scenario 2 where job.query end up with a stateful execution, JobId is populated and QueryId is not. I think this scenario is consistent with scenario 3 which makes sense.

It seems like the reason we are seeing jobId != null and queryId != null is in the case where a query can be stateless but setQueryPreviewEnabled is false which is covered in testStatelessQueries.

Please take a look.

shollyman (approved):

PR #3108, opened 2024-02-02
ci: unmanaged dependency check

suztomo:
b/320677879


suztomo:
It failed

> This pull request seems to add new third-party dependency, [org.threeten:threeten-extra, org.objenesis:objenesis, net.bytebuddy:byte-buddy-agent, org.mockito:mockito-core, org.apache.arrow:arrow-memory-core, org.apache.arrow:arrow-memory-netty, org.immutables:value, com.google.flatbuffers:flatbuffers-java, org.apache.arrow:arrow-vector, org.apache.arrow:arrow-format], among the artifacts listed in ./pom.xml.

PR #3133, opened 2024-02-09
feat: add MetadataCacheStatistics to Job QueryStatistics

### LINKED ISSUE #3134: Add support for metadata cache statistics in query statistics

PhongChuong:
Note: This only includes IT test for when the external table does not have metadata cache enabled. Since we are moving away from the GCP java-docs-samples-testing, the IT for when metadata cache enabled with be added after the migration due to the need to create BigLake table with external connection.

Fixes #3134 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
@@ -358,9 +358,7 @@ public abstract class JobStatistics implements Serializable {
     private final List<TimelineSample> timeline;
     private final Schema schema;
     private final SearchStats searchStats;
-
     private final MetadataCacheStats metadataCacheStats;
-
     private final List<QueryParameter> queryParameters;
 
     /**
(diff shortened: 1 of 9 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
```diff
@@ -761,6 +774,11 @@ public SearchStats getSearchStats() {
       return searchStats;
     }
 
+    /** Statistics for metadata caching in BigLake tables. */
```
shollyman, line 777: nit:  possibly include a link to more information about metadata caching, e.g. https://cloud.google.com/bigquery/docs/biglake-intro or elsewhere?

shollyman (approved):

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
PhongChuong: Added to comment of MetadataCacheStats class.
Thanks for the review.

PR #3236, opened 2024-04-10
feat: Add Range object to allow reading range value

### LINKED ISSUE #3235: Support reading range value

PhongChuong:
This PR also adds the ability to use Range query parameter

Fixes #3235 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
@@ -67,7 +67,10 @@ public class FieldValue implements Serializable {
     REPEATED,
 
     /** A {@code FieldValue} for a field of type {@link LegacySQLTypeName#RECORD}. */
-    RECORD
+    RECORD,
+
+    /** A {@code FieldValue} for a field of type {@link LegacySQLTypeName#RANGE}. */
+    RANGE
   }
 
   private FieldValue(Attribute attribute, Object value) {
@@ -239,6 +242,10 @@ public class FieldValue implements Serializable {
    */
   @SuppressWarnings("unchecked")
   public Range getRangeValue() {
+    if (attribute == Attribute.RANGE) {
+      return (Range) value;
+    }
+    // Provide best effort to convert value to Range object.
     return Range.of(getStringValue());
   }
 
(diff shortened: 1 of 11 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Range.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.Serializable;
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class Range implements Serializable {
+  private static final long serialVersionUID = 1L;
+
+  /** Returns the start value of the range. A null value represents an unbounded start. */
+  @Nullable
+  public abstract String getStart();
```
shollyman, line 30: A usability question:  currently the Range object returns the string representation, which feels limiting.  Can we do something like have the getters return FieldValue?  That would make it possible for users to do things like range.getStart().getTimestampInstant() for example with a `RANGE<TIMESTAMP>`.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Range.java
```diff
@@ -0,0 +1,87 @@
...
+    return new AutoValue_Range.Builder();
+  }
+
+  /**
+   * Creates an instance of {@code Range} from a string representation.
+   *
+   * <p>The expected string format is: "[start, end)", where start and end are string format of
+   * [DATE, TIME, TIMESTAMP].
+   */
+  public static Range of(String value) throws IllegalArgumentException {
```
shollyman, line 64: A question here I'm still wrestling with as well is what formats we parse.  `of()` here is clearly focused on the API representation.  Should we also support literal construction e.g. `RANGE<DATETIME> [UNBOUNDED, 2023-04-11 12:34:56)` ?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Range.java
PhongChuong: Updated the getters to return FieldValue. PTAL.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
```diff
@@ -229,6 +229,19 @@ public BigDecimal getNumericValue() {
...
+  /**
+   * Returns this field's value as a {@link Range}. This method should only be used * if the
+   * corresponding field has {@link LegacySQLTypeName#RANGE} type.
+   *
+   * @throws ClassCastException if the field is not a primitive type
+   * @throws IllegalArgumentException if the field's value could not be converted to {@link Range}
+   * @throws NullPointerException if {@link #isNull()} returns {@code true}
+   */
+  @SuppressWarnings("unchecked")
+  public Range getRangeValue() {
```
shollyman, line 241: How do you propagate the RangeElementType from the schema when constructing the range?  I _think_ range.of() should probably be a two-arg constructor (the inner type and string api representation).

PhongChuong: Updated to propagate RangeElementType when schema is available. Provides best effort when schema is not available. Also, updated Range to use FieldElementType instead of StandardSQLTypeName to be more consistent with QueryParameterValue.

shollyman (approved):
Looks good.  I got caught up misunderstanding the QueryParameterValue mappings until I realized java uses it to provide both value and typing information.

PhongChuong:
Thank you for the review.

PR #3255, opened 2024-04-19
feat: Add getStringOrDefault method to FieldValue

### LINKED ISSUE #3179: FieldValueList & FieldValue throwing null pointer exception when getting value of nullable field
I have a table with numerous nullable fields but when using the Java api to get the data from a row I've found that instead of simply getting null instead of a string the code throws an exception.

https://github.com/googleapis/java-bigquery/blob/45c5096c756a6599233935553708af4f53de5d58/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java#L107-L121

As users are likely to want to map all results from a query regardless of whether the field was null or not it would be better if the methods for getting a String or Long handled null gracefully. Or at least provide an alternative method where the caller can provide a default value or null if the fields value is null. 

Fortunately as I'm using Kotlin I can work around this for now with an extension function on FieldValueList:

```
fun FieldValueList.getNullable(name: String): FieldValue? = if (this.get(name).isNull) null else this.get(name)
```

which in turn means I can simply do `row.getNullable("my_nullable_column")?.stringValue` when mapping the _FieldValueList_ to an object.

If `getStringValue()` had on overloaded method that takes an arg that's used as a default I could have done something like:

```
row.get("my_nullable_column").getStringValue(null)
```

which would be convenient. Also this could be added without breaking the existing API.

PhongChuong:

Fixes #3179 ☕️

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,13 +60,13 @@ implementation 'com.google.cloud:google-cloud-bigquery'
 If you are using Gradle without BOM, add this to your dependencies:
 
 ```Groovy
-implementation 'com.google.cloud:google-cloud-bigquery:2.38.2'
+implementation 'com.google.cloud:google-cloud-bigquery:2.39.0'
 ```
 
 If you are using SBT, add this to your dependencies:
 
 ```Scala
-libraryDependencies += "com.google.cloud" % "google-cloud-bigquery" % "2.38.2"
+libraryDependencies += "com.google.cloud" % "google-cloud-bigquery" % "2.39.0"
 ```
 <!-- {x-version-update-end} -->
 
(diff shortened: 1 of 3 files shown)


shollyman (commented):
Should we exercise this in the tests using a null string rather than sentinel string value?  Similarly, do we need to do this for all the typed getters?

PhongChuong:
Added comment suggestions.
And as discussed offline, we will only provide getStringOrDefault for String types for now. Other types can be added as if there are user requests for it.

shollyman (approved):

PhongChuong:
Thanks for the review.

PR #3393, opened 2024-07-16
testing: reduce flakiness of tests

shollyman:
This PR addresses recent flakes observed in two tests.  In the first, we change validation for a DML test to examine DML stats rather then rescanning row data.  In the second, we switch a test query used for faster query to be deterministic, as the suspicion is that stale cached results are impacting assertions about how the query was run.

Fixes: https://github.com/googleapis/java-bigquery/issues/3386
Fixes: https://github.com/googleapis/java-bigquery/issues/3391


PhongChuong (approved):

PhongChuong:
Thanks for the fix.

PR #3394, opened 2024-07-17
docs: add simple query connection read api sample

### LINKED ISSUE #3404: Add code snippet for using Connection

PhongChuong:

Fixes #3404 ☕️


samples/snippets/src/main/java/com/example/bigquery/SimpleQueryConnectionReadApi.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.cloud.bigquery.BigQueryResult;
+import com.google.cloud.bigquery.Connection;
+import com.google.cloud.bigquery.ConnectionSettings;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public class SimpleQueryConnectionReadApi {
+
+  public static void main(String[] args) {
+    // TODO(developer): Replace this query before running the sample.
```
shollyman, line 33: Given the example has dependencies on the output column names, this TODO seems unnecessary.

If we want a more complete example, should we have a query example that dynamically walks result schema to print the output?  Likely a different sample, but something to consider.

shollyman (approved):

PhongChuong:
Thanks for the review.

samples/snippets/src/main/java/com/example/bigquery/SimpleQueryConnectionReadApi.java
PhongChuong: Removed the unnecessary TODO. Lets discuss offline what additional samples would be useful.

PR #3398, opened 2024-07-23
feat: Add ability to specify RetryOptions and BigQueryRetryConfig when create job and waitFor

### LINKED ISSUE #3399: Add ability to specify retry options for job create

PhongChuong:
Add ability to specify RetryOptions and BigQueryRetryConfig when creating jobs. The RetryOptions and BigQueryRetryConfig are embedded in JobOptions so the BigQuery API interface does not change. This can be extended to other options such as DatasetListOption, DatasetOption, etc. Those options will be added in a future PR as needed.

In addition:
This PR also overloads the job.waitFor() method to allow users to specify BigQueryRetryConfig.
This PR also has a minor refactor to JobTest.java.

Fixes #3399 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
@@ -197,12 +197,20 @@ public class Job extends JobInfo {
     return job == null || JobStatus.State.DONE.equals(job.getStatus().getState());
   }
 
+  /** See {@link #waitFor(BigQueryRetryConfig, RetryOption...)} */
+  public Job waitFor(RetryOption... waitOptions) throws InterruptedException {
+    return waitForInternal(DEFAULT_RETRY_CONFIG, waitOptions);
+  }
+
   /**
    * Blocks until this job completes its execution, either failing or succeeding. This method
    * returns current job's latest information. If the job no longer exists, this method returns
    * {@code null}. By default, the job status is checked using jittered exponential backoff with 1
    * second as an initial delay, 2.0 as a backoff factor, 1 minute as maximum delay between polls,
-   * 12 hours as a total timeout and unlimited number of attempts.
+   * 12 hours as a total timeout and unlimited number of attempts. For query jobs, the job status
+   * check can be configured to retry on specific BigQuery error messages using {@link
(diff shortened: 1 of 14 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -238,12 +239,23 @@ public boolean isDone() {
...
     checkNotDryRun("waitFor");
     Object completedJobResponse;
     if (getConfiguration().getType() == Type.QUERY) {
       completedJobResponse =
           waitForQueryResults(
               RetryOption.mergeToSettings(DEFAULT_JOB_WAIT_SETTINGS, waitOptions),
+              bigQueryRetryConfig,
               DEFAULT_QUERY_WAIT_OPTIONS);
     } else {
       completedJobResponse =
```
shollyman, line 261: Does this change need to plumb the new retry config down through the non-query path (`waitForJob`) as well?

shollyman (commented):
Generally looks good, just a question about this new configurability in the face of non-query jobs.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
PhongChuong: The current behavior for non-query path does not retry on BigQuery error messages so it does not need to be plumb through. I think we can consider adding retry on BigQuery error messages for non-query path also but it would change the existing default behavior so lets discuss offline and add it on a later PR if needed.

I've updated the doc string to make it more clear. PTAL.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -232,18 +241,46 @@ public boolean isDone() {
    * }
    * }</pre>
    *
+   * <p>Example usage of {@code waitFor()} with BigQuery retry configuration to retry on rate limit
+   * exceeded error messages for query jobs.
+   *
+   * <pre>{@code
+   * Job completedJob =
+   *     job.waitFor(
```
shollyman, line 249: I expect this will be a frequently referenced example, thanks for including it.

shollyman (approved):
Thanks for the clarifications, looks good!

PR #3460, opened 2024-08-27
Fix unnecessary boxed primitives in AutoValue classes

### LINKED ISSUE #3459: google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java does unnecessary boxing
This line is unnecessarily using a boxed `Long` instead of a `long`:

https://github.com/googleapis/java-bigquery/blob/421c8250de28b33fb5d2e175b88f1042194f8dc2/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java#L42

Future version of AutoValue report a diagnostic for this issue: https://github.com/google/auto/commit/328a25c161b4423ff71b38054d3f2502da394d37

cushon:
Fixes

```
google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java:42: error: [AutoValueUnnecessaryBoxing] property method com.google.cloud.bigquery.TableResult.getTotalRows() is primitive but parameter of setter method is not
    public abstract TableResult.Builder setTotalRows(Long totalRows);
                                        ^
```

See https://github.com/google/auto/commit/328a25c161b4423ff71b38054d3f2502da394d37

Fixes #3459 ☕️

If you write sample code, please follow the [samples format](
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).


PhongChuong:
Generally looks good.
Can you add an exception for the change in:
google-cloud-bigquery/clirr-ignored-differences.xml
And fix the conventionalcommits check.

cushon:
> Generally looks good. Can you add an exception for the change in: google-cloud-bigquery/clirr-ignored-differences.xml And fix the conventionalcommits check.

Thanks, done!

PhongChuong:
@cushon , it seems like ci / clirr is still failing.
Sorry for the late reply.

cpovirk:
@cushon is out for a couple more days, but I've been wondering: Would it make sense to leave the old method in place but as `@Deprecated` method that forwards to the new one? This PR isn't dangerous for anyone who recompiles against the new BigQuery version before using it, but it's dangerous for anyone who has previously compiled against BigQuery and then tries to use that precompiled jar with a new version of BigQuery at runtime. Given that, this PR would in principle justify a major-version bump, but I don't know how big the danger is in practice.

cushon:
> Would it make sense to leave the old method in place but as `@Deprecated` method that forwards to the new one?

Thanks, I'm going to try that out, I pushed another commit.

cushon:
I'm still getting a clirr error, any ideas what why the exemption isn't sufficient?

```
Error:  Unable to locate enclosing class com.google.cloud.bigquery. for nested class com.google.cloud.bigquery.$AutoValue_Annotations
Error:  7013: com.google.cloud.bigquery.TableResult$Builder: Abstract method 'public com.google.cloud.bigquery.TableResult$Builder setTotalRows(long)' has been added
```

google-cloud-bigquery/clirr-ignored-differences.xml
```diff
@@ -129,4 +129,9 @@
     <className>com/google/cloud/bigquery/StandardTableDefinition*</className>
     <method>*BigLakeConfiguration(*)</method>
   </difference>
-  </differences>
\ No newline at end of file
+  <difference>
+    <differenceType>7005</differenceType>
```
cpovirk, line 133:
```suggestion
    <differenceType>7013</differenceType>
```

It looks like it's now complaining about a new `abstract` method, I assume on the theory that subclasses would have to implement it. Given that the only subclass here will be the one generated by AutoValue (I hope :)), we should be able to suppress that, too.

Since the old method continues to exist in this iteration of the PR, that suppression can replace the old one from the time of the method removal. With that change, `JOB_TYPE=clirr .kokoro/build.sh` appears to pass.

Issue #2234, opened 2022-08-16
Support credentials from InputStream with BigQuery Read API (Connection.executeSelect)


prash-mi:
`ExecuteSelect` errors out while initialising connection with the read API when `GOOGLE_APPLICATION_CREDENTIALS` is not set and the credential are passed using `setCredentials(credentials)` instead. 

Expected behaviour: Connection with Read API should be initialised if  credentials are being passed using the mentioned method. 

Details:

SampleCode: 
```

      TableId targetTable = TableId.of(projectId, defaultDataset, UUID.randomUUID().toString());
        GoogleCredentials credentials;

        try (FileInputStream serviceAccountStream = new FileInputStream(keyFile)) {
            credentials = ServiceAccountCredentials.fromStream(serviceAccountStream);
        }

        // Instantiate a BigQuery Client
        BigQuery bigquery =
                BigQueryOptions.newBuilder()
                        .setCredentials(credentials)
                        .setProjectId(projectId)
                        .build()
                        .getService();


        ConnectionSettings connectionSettings =
                ConnectionSettings.newBuilder()
                        .setDestinationTable(targetTable)
                        .build();

        Connection connection = bigquery.createConnection(connectionSettings);
        String query = "SELECT * FROM `a.b.c`";
        BigQueryResult bigQueryResult = connection.executeSelect(query);

```

Exception: 

```
com.google.cloud.bigquery.BigQuerySQLException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
    at com.google.cloud.bigquery.ConnectionImpl.executeSelect(ConnectionImpl.java:185) ~[google-cloud-bigquery-2.11.0.jar!/:2.11.0]
```

PR #3465, opened 2024-08-29
fix: executeSelect now use provided credentials instead of GOOGLE_APP…

### LINKED ISSUE #2531: BigQueryStorage: Connection API uses GOOGLE_APPLICATION_CREDENTIALS instead of credentials configured on instance.

#### Environment details

1. Specify the API at the beginning of the title. For example, "BigQuery: ...").
   General, Core, and Other are also allowed as types
2. OS type and version: Mac Ventura / Linux Ubuntu / 
3. Java version: 17 adoptium 
4.  version(s): 2.22.0 2.19.1

#### Steps to reproduce

  1. create client with non-default credentials.
  2. run query using  that takes longer than 10 seconds so that it falls out of the fast query pathway
  3. 💥 

#### Code: 
```java
// don't have the GOOGLE_APPLICATION_CREDENTIALS env variable set.
BigQueryCredentials credentials = ServiceAccountCredentials.fromStream(inStream)

BigQueryOptions bigQueryOptions = BigQueryOptions.newBuilder()
    .setProjectId(projectId)
    .setCredentials(credentials)
            .setLocation(region)
            .build()
Bigquery bigquery = bigQueryOptions.getService()
ConneptionSettings connectionSettings = ConnectionSettings.newBuilder()
       .setUseReadAPI(false)
       .setRequestTimeout(10L)
       .setMaxResults(100L)
       .setUseQueryCache(true)
       .build();
Connection connection = bigquery.createConnection(connectionSettings);
String selectQuery = "<query that takes a little while and moves past the initial jobs.query timeout";
ListenableFuture<ExecuteSelectResponse> executeSelectFuture = connection.executeSelectAsync(selectQuery, ...);
```

#### Stack trace
```
The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

#### External references such as API reference guides

If the highThroughPut read path is hit, a BigQueryReadClient is created using the default `BigQueryReadClient.create()` method. [here](https://github.com/googleapis/java-bigquery/blob/v2.22.0/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java#L951). 

Because no credentials are passed, this looks for the `GOOGLE_APPLICATION_CREDENTIALS` instead of passing along the credentials that are configured for the bigquery client being used by the Connection.

PhongChuong:
…LICATION_CREDENTIALS

Fixes #2531, #2234 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
@@ -21,6 +21,7 @@ import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 
 import com.google.api.core.BetaApi;
 import com.google.api.core.InternalApi;
+import com.google.api.gax.core.FixedCredentialsProvider;
 import com.google.api.services.bigquery.model.GetQueryResultsResponse;
 import com.google.api.services.bigquery.model.JobConfigurationQuery;
 import com.google.api.services.bigquery.model.QueryParameter;
@@ -35,6 +36,7 @@ import com.google.cloud.bigquery.spi.v2.BigQueryRpc;
 import com.google.cloud.bigquery.storage.v1.ArrowRecordBatch;
 import com.google.cloud.bigquery.storage.v1.ArrowSchema;
 import com.google.cloud.bigquery.storage.v1.BigQueryReadClient;
+import com.google.cloud.bigquery.storage.v1.BigQueryReadSettings;
 import com.google.cloud.bigquery.storage.v1.CreateReadSessionRequest;
 import com.google.cloud.bigquery.storage.v1.DataFormat;
 import com.google.cloud.bigquery.storage.v1.ReadRowsRequest;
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
```diff
@@ -955,7 +957,12 @@ BigQueryResult highThroughPutRead(
 
     try {
       if (bqReadClient == null) { // if the read client isn't already initialized. Not thread safe.
-        bqReadClient = BigQueryReadClient.create();
+        BigQueryReadSettings settings =
+            BigQueryReadSettings.newBuilder()
+                .setCredentialsProvider(
+                    FixedCredentialsProvider.create(bigQueryOptions.getCredentials()))
```
shollyman, line 963: Is the FixedCredentialsProvider indirection needed, or can we set the retrieved credentials directly?  Is this a typing issue, or does the indirection get us some other benefit?

shollyman (approved):
one minor question about how the credentials are propagated, otherwise LGTM

PhongChuong:
Thanks for the review.

PR #3499, opened 2024-09-26
feat: add max staleness to ExternalTableDefinition

PhongChuong:

Fixes #<issue_number_goes_here> ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
@@ -320,7 +320,8 @@ public abstract class ExternalTableDefinition extends TableDefinition {
   abstract String getMetadataCacheModeInner();
 
   /**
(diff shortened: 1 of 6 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
```diff
@@ -351,6 +381,9 @@ public HivePartitioningOptions getHivePartitioningOptions() {
   com.google.api.services.bigquery.model.Table toPb() {
     Table tablePb = super.toPb();
     tablePb.setExternalDataConfiguration(toExternalDataConfigurationPb());
+    if (getMaxStaleness() != null) {
```
whuffman36, line 384: Does getMaxStaleness() construct a String or just return a reference? If it is constructing, it might be a good idea to store the result in a variable instead of calling the function twice in a row in the case that getMaxStaleness() does not return null. I don't think it matters much if the function just returns a reference.

whuffman36 (approved):
LGTM

PhongChuong:
Thanks for the review.

PR #3534, opened 2024-10-11
docs: Update SimpleApp to explicitly set project id

PhongChuong:
When running in Google Cloud console, the default google cloud project might not be set leading to 404 errors. This update requires the reader to explicitly set the project value.

This also removes the explicit use of UUID to set jobId as the library now internally generates the value.

If you write sample code, please follow the [samples format](
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).


whuffman36 (approved):
LGTM

PhongChuong:
Thanks for the review.

PR #3555, opened 2024-10-28
Enable maxTimeTravelHours in BigQuery java client library

### LINKED ISSUE #3538: Can we get the configured time travel window for a Table
Hi there,

I don't seem to find a way to get the configured time travel window for a Table (mentioned in the docs here https://cloud.google.com/bigquery/docs/time-travel#configure_the_time_travel_window), is that available from the java client or the API?

I've only found TableInfo.getNumTimeTravelPhysicalBytes() https://github.com/googleapis/java-bigquery/blob/e78e78cba7769a55b3e5937cf6a5a84f5259d116/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java#L555 and TableInfo.getCreationTime()

Thanks!
Robin

whuffman36:
The BigQuery API contains a field called [maxTimeTravelHours](https://github.com/googleapis/google-api-java-client-services/blob/main/clients/google-api-services-bigquery/v2/2.0.0/com/google/api/services/bigquery/model/Dataset.java#L238) that refers to the [Time Travel Window](https://cloud.google.com/bigquery/docs/time-travel#configure_the_time_travel_window) feature. This feature allows users to query deleted or updated data for an amount of time specified in the maxTimeTravelHours parameter. This change exposes the maxTimeTravelHours to users of the client library so they can set and update the amount of time that deleted or updated data is stored for a given dataset.

[Design Doc One Pager](https://docs.google.com/document/d/1hS9-8pZuvjeaCOlEmzvj928BN9Gvt8mXXuFeZMgvx34/edit?tab=t.0#heading=h.kyrewiptmx5v)

Fixes #3538 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DatasetInfo.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DatasetInfo.java
@@ -75,6 +75,7 @@ public class DatasetInfo implements Serializable {
   private final String defaultCollation;
   private final ExternalDatasetReference externalDatasetReference;
   private final String storageBillingModel;
+  private final Long maxTimeTravelHours;
 
   /** A builder for {@code DatasetInfo} objects. */
   public abstract static class Builder {
@@ -143,6 +144,13 @@ public class DatasetInfo implements Serializable {
     public abstract Builder setStorageBillingModel(String storageBillingModel);
 
     /**
+     *
+     * Optional. Amount of time in hours that deleted or updated data will remain accessible
+     * to be queried for all tables in the dataset. Default is 168 hours (7 days).
+     */
+    public abstract Builder setMaxTimeTravelHours(Long maxTimeTravelHours);
+
+    /**
      * The default encryption key for all tables in the dataset. Once this property is set, all
      * newly-created partitioned tables in the dataset will have encryption key set to this value,
      * unless table creation request (or query) overrides the key.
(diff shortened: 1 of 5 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DatasetInfo.java
```diff
@@ -142,6 +143,13 @@ public abstract Builder setExternalDatasetReference(
      */
     public abstract Builder setStorageBillingModel(String storageBillingModel);
 
+    /**
+     *
+     * Optional. Amount of time in hours that deleted or updated data will remain accessible
```
PhongChuong, line 148: Can you verify the description? I'm seeing slightly different description in the REST discovery document.

PhongChuong (approved):
Looks good. Just a minor comment regarding a method description.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DatasetInfo.java
whuffman36: Yes good idea, I've copied over the description from [API ](https://github.com/googleapis/google-api-java-client-services/blob/main/clients/google-api-services-bigquery/v2/2.0.0/com/google/api/services/bigquery/model/Dataset.java#L238)

PR #3586, opened 2024-11-19
feat: introduce `java.time` methods and variables

diegomarquezp:
This PR introduces `java.time` alternatives to existing `org.threeten.bp.*` methods, as well as switching internal variables (if any) to `java.time`

The main constraint is to keep the changes backwards compatible, so for each existing threeten method "`method1(org.threeten.bp.Duration)`" we will add an alternative with a _Duration_ (or _Timestamp_ when applicable) suffix: "`method1Duration(java.time.Duration)`".

For most cases, the implementation will be held in the `java.time` method and the old threeten method will just delegate the call to it. However, for the case of abstract classes, the implementation will be kept in the threeten method to avoid breaking changes (i.e. users that already overloaded the method in their user code).

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
@@ -16,11 +16,12 @@
 
 package com.google.cloud.bigquery;
 
-import static org.threeten.bp.temporal.ChronoField.HOUR_OF_DAY;
-import static org.threeten.bp.temporal.ChronoField.MINUTE_OF_HOUR;
-import static org.threeten.bp.temporal.ChronoField.NANO_OF_SECOND;
-import static org.threeten.bp.temporal.ChronoField.SECOND_OF_MINUTE;
+import static java.time.temporal.ChronoField.HOUR_OF_DAY;
+import static java.time.temporal.ChronoField.MINUTE_OF_HOUR;
+import static java.time.temporal.ChronoField.NANO_OF_SECOND;
+import static java.time.temporal.ChronoField.SECOND_OF_MINUTE;
 
+import com.google.api.core.ObsoleteApi;
 import com.google.api.services.bigquery.model.QueryParameterType;
 import com.google.api.services.bigquery.model.RangeValue;
 import com.google.auto.value.AutoValue;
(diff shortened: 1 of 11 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -345,7 +346,11 @@ public static QueryParameterValue interval(String value) {
     return of(value, StandardSQLTypeName.INTERVAL);
   }
 
-  /** Creates a {@code QueryParameterValue} object with a type of INTERVAL. */
+  /**
+   * Creates a {@code QueryParameterValue} object with a type of INTERVAL. This method is obsolete.
+   * Use {@link #interval(String)} instead.
+   */
+  @ObsoleteApi("Use interval(String) instead")
   public static QueryParameterValue interval(PeriodDuration value) {
```
diegomarquezp, line 354: @lqiu96 I'm not sure if this makes sense. [`PeriodDuration`](https://www.threeten.org/threeten-extra/apidocs/org.threeten.extra/org/threeten/extra/PeriodDuration.html) is part of Threeten Extra which only accepts `java.time` objects.
I'm guessing we may want to keep it since it's just a helper class that _happens_ to be from ThreeTen but not meant to be a backport (since this is not available in the JDK).

lqiu96: Hmm. Ideally, I think we should try to migrate away from threeten-extra as well. I'm guessing from your comment that PeriodDuration doesn't exist in the JDK so we can't build a 1:1 method signature replacement with the available JDK alternative.

If `interval(String)` has the same behavior, then I think that probably can use that as the alternative. But would be a question for the Bigquery SMEs to confirm that they have the same behavior.

diegomarquezp: Thank you. @PhongChuong can you please confirm whether marking this `threeten-extra`-based method as `@ObsoleteApi` makes sense? And does `interval(String)` as the suggested alternative make sense?

PhongChuong: @diegomarquezp , marking it as @ObsoleteApi and the suggested alternative makes sense.

diegomarquezp: Thanks!

PhongChuong (approved):

lqiu96 (approved):

PR #3602, opened 2024-12-09
feat(bigquery): support IAM conditions in datasets in Java client.

whuffman36:
This PR enables the use of IAM conditions when accessing a dataset through the create, update, and get API endpoints. The following changes are necessary:

- Add in a new `Expr` type in the `Acl` class for expressing the access conditions
- Create a new `DatasetOption` field to set the access policy version
- Set the access policy version before sending the rpc to the backend when applicable

Buganizer link: b/374156746

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
@@ -590,6 +590,68 @@ public final class Acl implements Serializable {
      */
     private final String location;
 
+    private static final long serialVersionUID = 7358264726377291156L;
+
+    static final class Builder {
+      private String expression;
+      private String title;
+      private String description;
+      private String location;
+
+      Builder() {}
+
+      Builder(Expr expr) {
+        this.expression = expr.expression;
+        this.title = expr.title;
+        this.description = expr.description;
+        this.location = expr.location;
+      }
+
+      Builder(com.google.api.services.bigquery.model.Expr bqExpr) {
+        this.expression = bqExpr.getExpression();
+        if (bqExpr.getTitle() != null) {
+          this.title = bqExpr.getTitle();
+        }
+        if (bqExpr.getDescription() != null) {
+          this.description = bqExpr.getDescription();
+        }
+        if (bqExpr.getLocation() != null) {
+          this.location = bqExpr.getLocation();
+        }
+      }
+
+      public Builder setExpression(String expression) {
+        this.expression = expression;
+        return this;
+      }
+
+      public Builder setTitle(String title) {
+        this.title = title;
+        return this;
+      }
+
+      public Builder setDescription(String description) {
+        this.description = description;
+        return this;
+      }
+
+      public Builder setLocation(String location) {
+        this.location = location;
+        return this;
+      }
+
+      public Expr build() {
+        return new Expr(this);
+      }
+    }
+
+    public Expr(Builder builder) {
+      this.expression = builder.expression;
+      this.title = builder.title;
+      this.description = builder.description;
+      this.location = builder.location;
+    }
+
     public Expr(String expression, String title, String description, String location) {
       this.expression = expression;
       this.title = title;
@@ -608,8 +670,11 @@ public final class Acl implements Serializable {
     }
 
     static Expr fromPb(com.google.api.services.bigquery.model.Expr bqExpr) {
-      return new Expr(
-          bqExpr.getExpression(), bqExpr.getTitle(), bqExpr.getDescription(), bqExpr.getLocation());
+      return new Builder(bqExpr).build();
+    }
+
+    public Builder toBuilder() {
+      return new Builder(this);
     }
 
     @Override
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
@@ -631,6 +631,11 @@ public final class Acl implements Serializable {
           && Objects.equals(this.description, other.description)
           && Objects.equals(this.location, other.location);
     }
+
+    @Override
+    public String toString() {
+      return toPb().toString();
+    }
   }
 
   private Acl(Entity entity, Role role) {
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
@@ -21,9 +21,9 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.api.core.ApiFunction;
 import com.google.api.services.bigquery.model.Dataset.Access;
 import com.google.api.services.bigquery.model.DatasetAccessEntry;
+import com.google.api.services.bigquery.model.Expr;
 import com.google.cloud.StringEnumType;
 import com.google.cloud.StringEnumValue;
-import com.google.api.services.bigquery.model.Expr;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Objects;
@@ -570,25 +570,23 @@ public final class Acl implements Serializable {
     }
   }
 
-  /**
-   * Expr represents the conditional information related to dataset access policies.
-   */
+  /** Expr represents the conditional information related to dataset access policies. */
   public static final class Expr implements Serializable {
     // Textual representation of an expression in Common Expression Language syntax.
     private final String expression;
     /**
-     * Optional. Title for the expression, i.e. a short string describing
-     * its purpose. This can be used e.g. in UIs which allow to enter the expression.
+     * Optional. Title for the expression, i.e. a short string describing its purpose. This can be
+     * used e.g. in UIs which allow to enter the expression.
      */
     private final String title;
     /**
-     * Optional. Description of the expression. This is a longer text which
-     * describes the expression, e.g. when hovered over it in a UI.
+     * Optional. Description of the expression. This is a longer text which describes the
+     * expression, e.g. when hovered over it in a UI.
      */
     private final String description;
     /**
-     * Optional. String indicating the location of the expression for error
-     * reporting, e.g. a file name and a position in the file.
+     * Optional. String indicating the location of the expression for error reporting, e.g. a file
+     * name and a position in the file.
      */
     private final String location;
 
@@ -600,7 +598,8 @@ public final class Acl implements Serializable {
     }
 
     com.google.api.services.bigquery.model.Expr toPb() {
-      com.google.api.services.bigquery.model.Expr bqExpr = new com.google.api.services.bigquery.model.Expr();
+      com.google.api.services.bigquery.model.Expr bqExpr =
+          new com.google.api.services.bigquery.model.Expr();
       bqExpr.setExpression(this.expression);
       bqExpr.setTitle(this.title);
       bqExpr.setDescription(this.description);
@@ -609,7 +608,8 @@ public final class Acl implements Serializable {
     }
 
     static Expr fromPb(com.google.api.services.bigquery.model.Expr bqExpr) {
-      return new Expr(bqExpr.getExpression(), bqExpr.getTitle(), bqExpr.getDescription(), bqExpr.getLocation());
+      return new Expr(
+          bqExpr.getExpression(), bqExpr.getTitle(), bqExpr.getDescription(), bqExpr.getLocation());
     }
 
     static Expr defaultExpr() {
@@ -620,6 +620,7 @@ public final class Acl implements Serializable {
     public int hashCode() {
       return Objects.hash(expression, title, description, location);
     }
+
     @Override
     public boolean equals(Object obj) {
       if (this == obj) {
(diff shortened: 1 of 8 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
```diff
@@ -568,9 +570,77 @@ Access toPb() {
...
+      bqExpr.setLocation(this.location);
+      return bqExpr;
+    }
+
+    static Expr fromPb(com.google.api.services.bigquery.model.Expr bqExpr) {
+      return new Expr(
+          bqExpr.getExpression(), bqExpr.getTitle(), bqExpr.getDescription(), bqExpr.getLocation());
+    }
+
+    @Override
```
shollyman, line 615: Include toString() as well as the normal methods like equals and hashcode?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -130,6 +130,21 @@ private void validateRPC() throws BigQueryException, IOException {
...
     try {
       validateRPC();
+      Integer accessPolicyVersion = null;
+      for (Map.Entry<Option, ?> entry : options.entrySet()) {
+        if (entry.getKey() == Option.ACCESS_POLICY_VERSION && entry.getValue() != null) {
+          accessPolicyVersion = (Integer) entry.getValue();
+        }
+      }
+      if (accessPolicyVersion != null) {
+        return bigquery
```
shollyman, line 140: nit:  I worry here is if there's a divergence between this chain and the default execute().  Is it worth doing a single chain of builders with a conditional?  Same with this pattern in the other RPCs (create, patch, etc).

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -1219,6 +1221,35 @@ public void testGetDatasetWithSelectedFields() {
     assertNull(dataset.getMaxTimeTravelHours());
   }
 
+  @Test
+  public void testGetDatasetWithAccessPolicyVersion() {
+    String accessPolicyDataset = RemoteBigQueryHelper.generateDatasetName();
+    User user = new User("liamhuffman@google.com");
```
shollyman, line 1227: Let's use a managed identity for this kind of thing rather than your own identity, particular for an IT test.  Same for the other tests that use your identity directly.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
whuffman36: Good point, updated to have toString() method.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
whuffman36: Yes that approach seems cleaner. I updated the conditional to toggle only the access policy version instead of creating an entirely new RPC.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
whuffman36: I've updated the user to use a default bigquery account for now. I found the account in my test project and I'm not sure if every test account follows the same naming pattern. We discussed a bit offline but this works as a workaround for the time being until we figure out how to integrate a managed account.

whuffman36: It appears that not all projects have the test account, as the integration tests failed due to the account not existing in the project. Looking into https://g3doc.corp.google.com/company/teams/eip-cloud/overground/user-guide/policies/org/iam.allowedPolicyMemberDomains.md?cl=head

whuffman36: Looks like using the default credentials works in this test project, it was just failing in my personal test project. The tests are using the default managed identity now.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -1285,6 +1318,49 @@ public void testUpdateDatasetWithSelectedFields() {
...
+            dataset
+                .toBuilder()
+                .setDescription("Updated Description")
+                .setLabels(null)
+                .setAcl(acls)
+                .build(),
+            datasetOption);
+    assertNotNull(updatedDataset);
+    assertEquals(updatedDataset.getDescription(), "Updated Description");
+    assertThat(updatedDataset.getLabels().isEmpty());
```
PhongChuong, line 1359: Currently, is there a way to validate the set the expr value?
I see a "condition" parameter in the REST discovery but I'm not sure if it is populated.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -289,6 +289,10 @@ public static DatasetOption fields(DatasetField... fields) {
       return new DatasetOption(
           BigQueryRpc.Option.FIELDS, Helper.selector(DatasetField.REQUIRED_FIELDS, fields));
     }
+
+    public static DatasetOption accessPolicyVersion(Integer accessPolicyVersion) {
```
PhongChuong, line 293: Can we get a doc string? In particular, the user would need to understand what the accessPolicyVersion integer values mean. A link to an article or public documentation is sufficient.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
```diff
@@ -568,9 +570,82 @@ Access toPb() {
     }
   }
 
+  /** Expr represents the conditional information related to dataset access policies. */
+  public static final class Expr implements Serializable {
```
PhongChuong, line 574: An addition of:
  private static final long serialVersionUID = somevalueL;
to be consistent.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
```diff
@@ -568,9 +570,82 @@ Access toPb() {
...
+     * expression, e.g. when hovered over it in a UI.
+     */
+    private final String description;
+    /**
+     * Optional. String indicating the location of the expression for error reporting, e.g. a file
+     * name and a position in the file.
+     */
+    private final String location;
+
+    public Expr(String expression, String title, String description, String location) {
```
PhongChuong, line 593: Are all the parameters needed to form an valid Expr?
E.g., Is it valid to for users to create an Expr with no description, location, title? (It seems like the answer is no from the parameter descriptions)
If that is the case, then providing a builder might be preferable.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -130,12 +130,23 @@ private void validateRPC() throws BigQueryException, IOException {
   public Dataset getDataset(String projectId, String datasetId, Map<Option, ?> options) {
     try {
       validateRPC();
-      return bigquery
-          .datasets()
-          .get(projectId, datasetId)
-          .setFields(Option.FIELDS.getString(options))
-          .setPrettyPrint(false)
-          .execute();
+      Integer accessPolicyVersion = null;
```
PhongChuong, line 133: Nit. Move declaration of accessPolicyVersion closer to usage.
E.g.,
```
      validateRPC();
      Bigquery.Datasets.Get bqGetRequest =
          bigquery
              .datasets()
              .get(projectId, datasetId)
              .setFields(Option.FIELDS.getString(options))
              .setPrettyPrint(false);
      for (Map.Entry<Option, ?> entry : options.entrySet()) {
        if (entry.getKey() == Option.ACCESS_POLICY_VERSION && entry.getValue() != null) {
          bqGetRequest.setAccessPolicyVersion((Integer) entry.getValue());
        }
      }
      return bqGetRequest.execute();
```

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -276,13 +297,23 @@ public boolean deleteDataset(String projectId, String datasetId, Map<Option, ?>
   public Dataset patch(Dataset dataset, Map<Option, ?> options) {
     try {
       validateRPC();
+      Integer accessPolicyVersion = null;
```
PhongChuong, line 300: Ditto regarding moving declaration of accessPolicyVersion closer to usage.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -174,12 +185,22 @@ public Tuple<String, Iterable<Dataset>> listDatasets(String projectId, Map<Optio
   public Dataset create(Dataset dataset, Map<Option, ?> options) {
     try {
       validateRPC();
-      return bigquery
-          .datasets()
-          .insert(dataset.getDatasetReference().getProjectId(), dataset)
-          .setPrettyPrint(false)
-          .setFields(Option.FIELDS.getString(options))
-          .execute();
+      Integer accessPolicyVersion = null;
```
PhongChuong, line 188: Ditto regarding moving declaration of accessPolicyVersion closer to usage.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -1219,6 +1221,37 @@ public void testGetDatasetWithSelectedFields() {
...
   }
 
+  @Test
+  public void testGetDatasetWithAccessPolicyVersion() throws IOException {
+    String accessPolicyDataset = RemoteBigQueryHelper.generateDatasetName();
+    ServiceAccountCredentials credentials =
+        (ServiceAccountCredentials) GoogleCredentials.getApplicationDefault();
+    User user = new User(credentials.getClientEmail());
+    Acl.Role role = Acl.Role.WRITER;
+    Acl.Expr condition =
```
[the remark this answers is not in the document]
whuffman36: Good point. Fixed to be after 01/01/2024

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -1219,6 +1221,37 @@ public void testGetDatasetWithSelectedFields() {
...
+    Dataset dataset =
+        bigquery.create(
+            DatasetInfo.newBuilder(accessPolicyDataset)
+                .setDescription("Some Description")
+                .setAcl(ImmutableList.of(acl))
+                .build(),
+            datasetOption);
+    assertThat(dataset).isNotNull();
+
+    Dataset datasetClone = bigquery.getDataset(accessPolicyDataset, datasetOption);
```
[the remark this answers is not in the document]
whuffman36: Fixed.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -1285,6 +1318,49 @@ public void testUpdateDatasetWithSelectedFields() {
...
+                .setDescription("Some Description")
+                .setLabels(Collections.singletonMap("a", "b"))
+                .build());
+    assertThat(dataset).isNotNull();
+
+    User user = new User(credentials.getClientEmail());
+    Acl.Role role = Acl.Role.WRITER;
+    Acl.Expr condition =
+        new Expr(
+            "request.time < timestamp('2030-01-01T00:00:00Z')",
```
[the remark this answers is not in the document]
whuffman36: Fixed.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
whuffman36: Tests have been updated to verify the remote datasets' conditions. The position of the returned remote dataset Acl containing the condition is not consistent, so the test has to iterate through all of the Acl objects to find the right one.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
whuffman36: Fixed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Acl.java
whuffman36: I think only the expression is required for a valid `Expr`. I have implemented a builder for the class but have kept the public constuctor for `Expr` for ease of use.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
whuffman36: Fixed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
whuffman36: Fixed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
whuffman36: Fixed.

PhongChuong (approved):
Thanks for the the changes.

PR #3624, opened 2025-01-06
fix(bigquery): Remove ReadAPI bypass in executeSelect()

whuffman36:
Currently, the ReadAPI is being bypassed in the executeSelect() function when requests are sent using fast query. This change makes the ReadAPI have higher precedence than fast query, bypassing jobs.query and instead using jobs.insert.

This change also updates queryRpc() to use the table data from the first page of the query results instead of the empty results of the unfinished job.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
@@ -234,7 +234,8 @@ class ConnectionImpl implements Connection {
     }
     try {
       // use jobs.query if possible
-      if (isFastQuerySupported()) {
+      // Using the ReadAPI should take precedence over fast query
+      if (isFastQuerySupported() && !connectionSettings.getUseReadAPI()) {
         logger.log(Level.INFO, "\n Using Fast Query Path");
         final String projectId = bigQueryOptions.getProjectId();
         final QueryRequest queryRequest =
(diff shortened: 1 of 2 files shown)


whuffman36:
Still working on ideas for a testing strategy to make sure the requests are going to the endpoints that we want, but I wanted to get your opinions on the fix.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
```diff
@@ -476,22 +476,29 @@ private BigQueryResult queryRpc(
     }
 
     // Query finished running and we can paginate all the results
-    if (results.getJobComplete() && results.getSchema() != null) {
+    // Results should be read using the high throughput read API if sufficiently large.
+    boolean shouldUseReadApi =
```
PhongChuong, line 480: nit: shouldUseReadApi is very similar to the method: useReadAPI which is used in getSubsequentQueryResultsWithJob. In addition, even when shouldUseReadApi == true here, the useReadAPI() can return false in the subsequent getSubsequentQueryResultsWithJob call. Perhaps, changing the name to a negative condition is sufficient. E.g., shouldNotUseReadApi (or something else better).

PhongChuong (commented):
Just a couple of comments; overall PR looks good.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
```diff
@@ -835,6 +837,16 @@ public Boolean getCacheHit() {
       return cacheHit;
     }
 
+    /** Returns whether the query result is read from the high throughput ReadAPI. */
+    public Boolean getUseReadApi() {
```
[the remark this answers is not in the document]
whuffman36: Done, I added the annotation to the setter method as well.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
whuffman36: What do you think of `resultsLargeEnoughForReadApi`?

PR #3650, opened 2025-01-25
feat: implement wasNull for BigQueryResultSet

### LINKED ISSUE #3648: ResultSet `getXXX` returns default value instead of `null` for nullable fields
ResultSet `getXXX` returns default value instead of `null` for nullable fields.

BigQueryResultImpl implementation of `getBoolean`, `getDouble`, `getLong` should return `null` the same as others methods

Image

Schema
Image

Data (See row 4)

Image

```Kotlin
private fun ResultSet.getFieldValue(field: Field): Any? {
    val fieldName = field.name
    return  when (field.type.standardType) {
            StandardSQLTypeName.BOOL -> getBoolean(fieldName)
            StandardSQLTypeName.INT64 -> getLong(fieldName)
            StandardSQLTypeName.FLOAT64 -> getDouble(fieldName)
            StandardSQLTypeName.NUMERIC, StandardSQLTypeName.BIGNUMERIC -> getBigDecimal(fieldName)
            StandardSQLTypeName.STRING -> getString(fieldName)
            StandardSQLTypeName.BYTES -> getBytes(fieldName)
            StandardSQLTypeName.TIMESTAMP -> getTimestamp(fieldName)?.toInstant()
            StandardSQLTypeName.DATE -> getDate(fieldName)?.toLocalDate()?.atStartOfDay()?.toInstant(ZoneOffset.UTC)
            StandardSQLTypeName.GEOGRAPHY -> getString(fieldName)?.toByteArray(StandardCharsets.UTF_8)
            StandardSQLTypeName.JSON -> getString(fieldName)
            StandardSQLTypeName.INTERVAL -> getObject(fieldName)
            StandardSQLTypeName.RANGE -> getObject(fieldName)
            else -> getObject(fieldName)
        }
}
```

PhongChuong:
- Implement wasNull as defined by java.sql.ResultSet for BigQueryResultSet.
- Add a new unittest file for BigQueryResultSet
- Add unittest for BigQueryResultSet getters for both FieldValue type buffer and read API type buffer
- Perform minor refactoring in BigQueryResultSet

Fixes #3648 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
@@ -666,7 +666,7 @@ public class BigQueryResultImpl<T> implements BigQueryResult<T> {
             TimeUnit.DAYS.toMillis(
                 Long.valueOf(
                     dateInt)); // For example int 18993 represents 2022-01-01, converting time to
-                               // milli seconds
+        // milli seconds
         return new Date(dateInMillis);
       }
     }
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
@@ -24,9 +24,7 @@ import java.sql.Time;
 import java.sql.Timestamp;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
-import java.time.ZoneId;
 import java.util.Map;
-import java.util.TimeZone;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


shollyman (commented):
This code seems fairly invasive with the hasNull plumbing, and I worry it'll be prone to error.  Could we move the hasNull state computation higher up into when we do cursor advancement?  Then it's just something like (pseudocode) `for field in row.fields { if field.isNull() { hasNull = true }}`?

shollyman:
> This code seems fairly invasive with the hasNull plumbing, and I worry it'll be prone to error. Could we move the hasNull state computation higher up into when we do cursor advancement? Then it's just something like (pseudocode) `for field in row.fields { if field.isNull() { hasNull = true }}`?

After reading more about java.sql.resultset, it seems my previous understanding was wrong. It's not whether the previous row had a null, its if the latest field fetched had a null, which does yield more complexity here.

PhongChuong:
Thanks for the reviews.

PR #3668, opened 2025-02-10
fix: retry ExceptionHandler not retrying on IOException

PhongChuong:
This PR fix an issue where the retry algorithm did not retrying based on the BIGQUERY_EXCEPTION_HANDLER configuration. This is caused by HttpBigQueryRpc translating IOExceptions into BigQueryException does not match the configurations set by BIGQUERY_EXCEPTION_HANDLER. To resolve this, we implement new internal methods (.*SkipExceptionTranslation) in HttpBigQueryRpc which do not translate the IOException. All retryable calls to HttpBigQueryRpc are updated to use these new methods.

In addition,
- To preserve existing BigQueryException structure, this PR updates BigQueryRetryHelper.runWithRetries to parse the IOExceptions thrown by the .*SkipExceptionTranslation methods and translate them into BigQueryRetryException. This is needed as the original HttpBigQueryRpc translate IOExceptions into BigQueryException.
- To preserve return values in the BigQuery interface for special cases such as HTTP_NOT_FOUND, this PR updates BigQueryImpl to checks the error code and return the appropriate BigQueryException.
- The unit test BigQueryImplTest.testCreateJobFailureShouldRetryExceptionHandlerExceptions verify that BIGQUERY_EXCEPTION_HANDLER exceptions (UnknownHostException & ConnectException) are being retried

Fixes b/394167052 ☕️

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -53,6 +53,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -271,8 +272,8 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
           runWithRetries(
               new Callable<com.google.api.services.bigquery.model.Dataset>() {
                 @Override
-                public com.google.api.services.bigquery.model.Dataset call() {
-                  return bigQueryRpc.create(datasetPb, optionsMap);
+                public com.google.api.services.bigquery.model.Dataset call() throws IOException {
+                  return bigQueryRpc.createSkipExceptionTranslation(datasetPb, optionsMap);
                 }
               },
               getOptions().getRetrySettings(),
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -509,9 +509,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
               getOptions().getClock(),
               EMPTY_RETRY_CONFIG);
-      // This null check is not strictly necessary as getDatasetSkipExceptionTranslation will never
-      // return null.
-      return answer == null ? null : Dataset.fromPb(this, answer);
+      return Dataset.fromPb(this, answer);
     } catch (BigQueryRetryHelperException e) {
       if (isRetryErrorCodeHttpNotFound(e)) {
         if (getOptions().getThrowNotFound()) {
@@ -865,9 +863,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
               getOptions().getClock(),
               EMPTY_RETRY_CONFIG);
-      // This null check is not strictly necessary as getTableSkipExceptionTranslation will never
-      // return null.
-      return answer == null ? null : Table.fromPb(this, answer);
+      return Table.fromPb(this, answer);
     } catch (BigQueryRetryHelperException e) {
       if (isRetryErrorCodeHttpNotFound(e)) {
         if (getOptions().getThrowNotFound()) {
@@ -909,9 +905,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
               getOptions().getClock(),
               EMPTY_RETRY_CONFIG);
-      // This null check is not strictly necessary as getModelSkipExceptionTranslation will never
-      // return null.
-      return answer == null ? null : Model.fromPb(this, answer);
+      return Model.fromPb(this, answer);
     } catch (BigQueryRetryHelperException e) {
       if (isRetryErrorCodeHttpNotFound(e)) {
         if (getOptions().getThrowNotFound()) {
@@ -953,9 +947,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
               getOptions().getClock(),
               EMPTY_RETRY_CONFIG);
-      // This null check is not strictly necessary as getRoutineSkipExceptionTranslation will never
-      // return null.
-      return answer == null ? null : Routine.fromPb(this, answer);
+      return Routine.fromPb(this, answer);
     } catch (BigQueryRetryHelperException e) {
       if (isRetryErrorCodeHttpNotFound(e)) {
         if (getOptions().getThrowNotFound()) {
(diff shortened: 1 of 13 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -1572,23 +1674,47 @@ private static com.google.api.services.bigquery.model.Job newJobPb() {
...
     assertThat(jobCapture.getValue().getJobReference().getJobId()).isEqualTo(id);
-    verify(bigqueryRpcMock).create(jobCapture.capture(), eq(EMPTY_RPC_OPTIONS));
+    verify(bigqueryRpcMock)
+        .createSkipExceptionTranslation(jobCapture.capture(), eq(EMPTY_RPC_OPTIONS));
   }
 
   @Test
-  public void testCreateJobFailureShouldRetry() {
-    when(bigqueryRpcMock.create(jobCapture.capture(), eq(EMPTY_RPC_OPTIONS)))
+  public void testCreateJobFailureShouldRetryExceptionHandlerExceptions() throws IOException {
```
shollyman, line 1694: Should we have similar testing for the TableDataWriteChannel that validates behavior for retryable IOException cases?  That one's potentially more tricky as it is focused more on inline uploads rather than streaming inserts, despite the poorly chosen name.

PhongChuong: Added retry tests to TableDataWriteChanelTest.
Good call out. TabelDataWriteChanel retry on SocketException which is not included in our BIGQUERY_EXCEPTION_HANDLER. I've added it to the general default retry behavior which is also used outside of TabelDataWriteChanel. I think this makes sense but we can move it into just have it within TabelDataWriteChanel if you think that is better.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -490,22 +497,28 @@ public Dataset getDataset(final DatasetId datasetId, DatasetOption... options) {
...
               getOptions().getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
-              getOptions().getClock());
-      if (getOptions().getThrowNotFound() && answer == null) {
-        throw new BigQueryException(HTTP_NOT_FOUND, "Dataset not found");
-      }
+              getOptions().getClock(),
+              EMPTY_RETRY_CONFIG);
+      // This null check is not strictly necessary as getDatasetSkipExceptionTranslation will never
+      // return null.
```
whuffman36, line 513: If null is not a valid state for `answer`, the code should reflect that. Could we just return `Dataset.fromPb(this, answer)`?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -826,12 +863,18 @@ public com.google.api.services.bigquery.model.Table call() {
...
               getOptions().getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
-              getOptions().getClock());
-      if (getOptions().getThrowNotFound() && answer == null) {
-        throw new BigQueryException(HTTP_NOT_FOUND, "Table not found");
-      }
+              getOptions().getClock(),
+              EMPTY_RETRY_CONFIG);
+      // This null check is not strictly necessary as getTableSkipExceptionTranslation will never
+      // return null.
```
whuffman36, line 869: Same comment about null as above

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -864,12 +907,18 @@ public com.google.api.services.bigquery.model.Model call() {
...
               getOptions().getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
-              getOptions().getClock());
-      if (getOptions().getThrowNotFound() && answer == null) {
-        throw new BigQueryException(HTTP_NOT_FOUND, "Model not found");
-      }
+              getOptions().getClock(),
+              EMPTY_RETRY_CONFIG);
+      // This null check is not strictly necessary as getModelSkipExceptionTranslation will never
+      // return null.
```
whuffman36, line 913: Same comment about null as above

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -902,12 +951,18 @@ public com.google.api.services.bigquery.model.Routine call() {
...
               getOptions().getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
-              getOptions().getClock());
-      if (getOptions().getThrowNotFound() && answer == null) {
-        throw new BigQueryException(HTTP_NOT_FOUND, "Routine not found");
-      }
+              getOptions().getClock(),
+              EMPTY_RETRY_CONFIG);
+      // This null check is not strictly necessary as getRoutineSkipExceptionTranslation will never
+      // return null.
```
whuffman36, line 957: Same comment about null as above

whuffman36, line 957: Same comment about null as above

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1258,12 +1322,18 @@ public com.google.api.services.bigquery.model.Job call() {
...
               getOptions().getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
-              getOptions().getClock());
-      if (getOptions().getThrowNotFound() && answer == null) {
-        throw new BigQueryException(HTTP_NOT_FOUND, "Job not found");
-      }
+              getOptions().getClock(),
+              EMPTY_RETRY_CONFIG);
+      // This null check is not strictly necessary as getJobSkipExceptionTranslation will never
+      // return null.
```
whuffman36, line 1328: Same comment about null as above

whuffman36 (approved):
LGTM! Nice work on such a massive PR

shollyman (approved):
One more minor callout for delete behavior, but otherwise LGTM.

Also, "SkipExceptionTranslation" may forever be burned into my retinas, but thanks for slogging through this refactor.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -672,18 +702,19 @@ public boolean delete(JobId jobId) {
...
                   completeJobId.getProject(), completeJobId.getJob(), completeJobId.getLocation());
             }
           },
           getOptions().getRetrySettings(),
           BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
-          getOptions().getClock());
-    } catch (RetryHelper.RetryHelperException e) {
+          getOptions().getClock(),
+          EMPTY_RETRY_CONFIG);
+    } catch (BigQueryRetryHelperException e) {
```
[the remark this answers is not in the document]
PhongChuong: The main reason is to preserve existing behavior. I.e., the existing behavior does not check for HTTP not found code. Semantically, I think the reason is because returning true/false is not a great option during a retry for delete when HTTP_NOT_FOUND occurs. In particular, we can get HTTP_NOT_FOUND from the following cases:
1. The job never exited. We should return the HTTP_NOT_FOUND bigquery exception.
2. The job was deleted in a previous retry attempt. We should return true.
However, there is no way to distinguish the difference between the 2 cases from the client side alone.

PhongChuong:
Thanks for the review. I've comment on the `isRetryErrorCodeHttpNotFound` check variation for delete job. PTAL

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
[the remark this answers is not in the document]
shollyman: These two points are valid for all resource deletions RPC.  

I'm fine with preservation of existing behavior, flawed as it may be.  We could also revisit this later if we choose, so consistency keeps the risk lower for this big refactor.

PR #3693, opened 2025-02-25
feat(bigquery): Implement getArray in BigQueryResultImpl

### LINKED ISSUE #3655: Implement getArray() in BigQueryResultImpl
Add support for array data types in BigQueryResultImpl.

Currently, there isn't a way to read repeated fields in BigQuery.

`getArray` method is not implemented and throws: 'java.sql.SQLFeatureNotSupportedException' exception.

whuffman36:

Fixes #3655 ☕️

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@62b2cac7ed8198b15735ed49ab1e5cf35480ba46 # v2.4.0
+        uses: ossf/scorecard-action@f49aabe0b5af0936a0987cfb85d86b75731b0186 # v2.4.1
         with:
           results_file: results.sarif
           results_format: sarif
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -59,7 +59,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         with:
           name: SARIF file
           path: results.sarif
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
@@ -113,7 +113,7 @@ public class BigQueryResultImpl<T> implements BigQueryResult<T> {
     return underlyingResultSet;
   }
 
-  public class BigQueryResultSet extends AbstractJdbcResultSet {
+  private class BigQueryResultSet extends AbstractJdbcResultSet {
     private boolean hasReachedEnd =
         false; // flag which will be set to true when we have encountered a EndOfStream or when
     // curTup.isLast(). Ref: https://github.com/googleapis/java-bigquery/issues/2033
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
@@ -16,14 +16,19 @@
 
 package com.google.cloud.bigquery;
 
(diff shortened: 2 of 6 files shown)


.github/workflows/scorecard.yml
whuffman36, line 1: These files changes were added automatically to the PR by a bot...not sure if I can revert them or if this is expected behavior, it seems like they are just updating some dependency versions.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryResultImpl.java
```diff
@@ -114,6 +119,77 @@ private class BigQueryResultSet extends AbstractJdbcResultSet {
...
+        } else if (array instanceof FieldValueList) {
+          this.array = new ArrayList<>((FieldValueList) array);
+        } else {
+          this.array = (List<Object>) array;
+        }
+      }
+
+      @Override
+      public String getBaseTypeName() throws SQLException {
+        throw new SQLFeatureNotSupportedException();
```
whuffman36, line 137: I left these as unimplemented for now, as this is the only place where this class is used, and these methods are not necessary. Lmk if you think these should be implemented for BigQueryArrayResult

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryResultImplTest.java
```diff
@@ -184,6 +211,7 @@ public void testResultSetReadApi() throws InterruptedException, SQLException {
     rowValues.put("timestamp", EXPECTED_TIMESTAMP.getTime() * 1000);
     rowValues.put("time", EXPECTED_TIME.getTime() * 1000);
     rowValues.put("date", DATE_INT);
+    rowValues.put("intArray", EXPECTED_INT_ARRAY);
```
PhongChuong, line 214: Missing EXPECTED_STRING_ARRAY?

whuffman36: Fixed.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -3925,6 +3925,76 @@ public void testExecuteSelectSinglePageTableRow() throws SQLException {
...
+    assertTrue(rs.next()); // second row
+    // second row is non null, comparing the values
+    assertEquals("StringValue1", rs.getString("StringField"));
+    assertTrue(rs.getDouble("BigNumericField") == 0.3333333333333333d);
+    assertFalse(rs.getBoolean("BooleanField"));
+    assertNotNull(rs.getBytes("BytesField"));
+    assertEquals(1, rs.getInt("IntegerField"));
+    assertEquals(1534680695123L, rs.getTimestamp("TimestampField").getTime());
+    assertEquals(
+        java.sql.Date.valueOf("2018-08-18").toLocalDate(), rs.getDate("DateField").toLocalDate());
```
whuffman36, line 3985: For whatever reason, when using the ReadApi, the `getDate()` function returns the date as local instead of UTC. The `Date` object is different when converted from `FieldValue`, like in `testExecuteSelectSinglePageTableRow()` as opposed to from `Integer` in this case when reading from Arrow. The raw epoch time values are the same, but the `FieldValue` conversion is in UTC format, while the `Integer` conversion is in local format.

The same is true for the `getTime()` function.

Screenshot to show the difference: https://screenshot.googleplex.com/6wGhSehFwRV6wdC

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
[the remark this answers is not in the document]
whuffman36: I've added a new IT test to test the ReadAPI path. Turns out that path had not been tested yet, and so the expected responses for `getDate()` and `getTime()` were inaccurate. The ReadAPI path returns the `Date` and `Time` objects in local format instead of UTC format.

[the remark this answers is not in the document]
whuffman36: As discussed offline, it looks more like my local machine had a different local time than that of the VM running the kokoro, resulting in different dates and time objects. This has been resolved.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
whuffman36: As discussed offline, it looks more like my local machine had a different local time than that of the VM running the kokoro, resulting in different dates and time objects. This has been resolved.

PhongChuong (approved):

Issue #3708, opened 2025-03-03
Update Dependency: io.netty.netty-common Vulnerability


radkomateusz:
Hi java-bigquery maintainers,

Could you please update the library dependencies to use a newer version of io.netty.netty-common that does not contain known vulnerabilities? This issue was identified some time ago, but it hasn't been addressed in the last few releases.

Details:
The current dependency version in google-cloud-bigquery (v2.48.1) includes netty-common with known security vulnerabilities:
🔗 [Maven Repository Link](https://mvnrepository.com/artifact/com.google.cloud/google-cloud-bigquery/2.48.1)

Would it be possible to update this dependency in an upcoming release?

Thanks for your help! 🙂

PhongChuong:
[v2.49.0](https://mvnrepository.com/artifact/com.google.cloud/google-cloud-bigquery/2.49.0) no longer has the io.netty.netty-common vulnerability.
Thank you for the bug report.

PR #3715, opened 2025-03-12
deps: exclude io.netty:netty-common from org.apache.arrow:arrow-memor…

PhongChuong:
…y-netty

Addresses #3708 ☕️

### DIFF
--- a/google-cloud-bigquery/pom.xml
+++ b/google-cloud-bigquery/pom.xml
@@ -116,6 +116,16 @@
     <dependency>
       <groupId>org.apache.arrow</groupId>
       <artifactId>arrow-memory-netty</artifactId>
+      <!--
+      Due to JDK8 requirement, we are forced to use arrow version below v18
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/pom.xml
```diff
@@ -116,6 +116,16 @@
...
       <groupId>org.apache.arrow</groupId>
       <artifactId>arrow-memory-netty</artifactId>
+      <!--
+      Due to JDK8 requirement, we are forced to use arrow version below v18
+      which has io.netty:netty-common a dep. and its vulnerabilities
+      -->
+      <exclusions>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty-common</artifactId>
```
suztomo, line 126: You need to re-declare the dependency with a good version.

image

From this example https://stackoverflow.com/questions/9119055/how-to-exclude-maven-dependencies

PR #3747, opened 2025-04-08
feat(bigquery): Integrate Otel in client lib

whuffman36:
Adds OpenTelemetry tracing into the BigQueryImpl class, enabling in the following methods:

- (Dataset, Table, Job, Routine) Create
- (Dataset, Table, Job, Routine, Model, IamPolicy) Get
- (Dataset, Table, Job, Routine, Model, IamPolicy) Update
- (Dataset, Table, Routine, Model) Delete
- List (Datasets, Tables, Jobs, Routines, Models)
- Job Cancel
- insertAll
- testIamMethods
- queryRpc
- getQueryResults

This is one PR of several to fully integrate OTel into the client library, broken up into chunks to make review easier.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -496,24 +496,24 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
         createException = e;
       }
 
-    if (!idRandom) {
-      if (createException instanceof BigQueryException
-          && createException.getCause() != null
-          && createException.getCause().getMessage() != null) {
+      if (!idRandom) {
+        if (createException instanceof BigQueryException
+            && createException.getCause() != null
+            && createException.getCause().getMessage() != null) {
 
-        Pattern pattern = Pattern.compile(".*Already.*Exists:.*Job.*", Pattern.CASE_INSENSITIVE);
-        Matcher matcher = pattern.matcher(createException.getCause().getMessage());
+          Pattern pattern = Pattern.compile(".*Already.*Exists:.*Job.*", Pattern.CASE_INSENSITIVE);
+          Matcher matcher = pattern.matcher(createException.getCause().getMessage());
 
-        if (matcher.find()) {
-          // If the Job ALREADY EXISTS, retrieve it.
-          Job job = this.getJob(jobInfo.getJobId(), JobOption.fields(JobField.STATISTICS));
+          if (matcher.find()) {
+            // If the Job ALREADY EXISTS, retrieve it.
+            Job job = this.getJob(jobInfo.getJobId(), JobOption.fields(JobField.STATISTICS));
 
-          long jobCreationTime = job.getStatistics().getCreationTime();
-          long jobMinStaleTime = System.currentTimeMillis();
-          long jobMaxStaleTime =
-              java.time.Instant.ofEpochMilli(jobMinStaleTime)
-                  .minus(1, java.time.temporal.ChronoUnit.DAYS)
-                  .toEpochMilli();
+            long jobCreationTime = job.getStatistics().getCreationTime();
+            long jobMinStaleTime = System.currentTimeMillis();
+            long jobMaxStaleTime =
+                java.time.Instant.ofEpochMilli(jobMinStaleTime)
+                    .minus(1, java.time.temporal.ChronoUnit.DAYS)
+                    .toEpochMilli();
 
             // Only return the job if it has been created in the past 24 hours.
             // This is assuming any job older than 24 hours is a valid duplicate JobID


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/OpenTelemetryHelper.java
```diff
@@ -0,0 +1,710 @@
...
+  }
+
+  public Span datasetCreateSpan(DatasetInfo datasetInfo, Map<BigQueryRpc.Option, ?> optionsMap) {
+    if (!enabled) {
+      return null;
+    }
+    Span span =
+        tracer
+            .spanBuilder("datasetCreate")
+            .setAttribute("datasetId", datasetInfo.getDatasetId().getDataset())
```
PhongChuong, line 50: I'm not a big fan of the OpenTelemetryHelper class (or helper class in general). IMO, having the helper class setAttribute based on datasetInfo.get.*() is that when adding new fields to DatasetInfo, it is not immediately clear that OpenTelemetryHelper has to be updated also.

Consider having a DatasetInfo.getOtelAttributes() which returns [Attributes](https://github.com/open-telemetry/opentelemetry-java/blob/29523e62ea351ed113a1ae060cc549c8920e9b47/api/all/src/main/java/io/opentelemetry/api/common/Attributes.java#L34). Then setting the span by the [setAllAttributes](https://github.com/open-telemetry/opentelemetry-java/blob/29523e62ea351ed113a1ae060cc549c8920e9b47/api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java#L263) method. Similarly, an helper method can be added to BigQueryImpl.java similar to the static helper [optionMap](https://github.com/googleapis/java-bigquery/blob/3df4f2745ee745f717eeeab56bdbe9263d2649e0/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java#L1708).

So in BigQueryImpl.java, it would look something like this:
```
  public Dataset create(DatasetInfo datasetInfo, DatasetOption... options) {
  ...
    final Span datasetCreateSpan = null;
    if (getOptions().getOpenTelemetryTracer()) {
      datasetCreateSpan = getOptions().getOpenTelemetryTracer().spanBuilder("datasetCreate")
          .setAllAttributes(datasetInfo.getOtelAttributes())
          .setAllAttributes(otelAttributesFromOptions(options))
          .startSpan();
    }
    try (Scope datasetCreateScope = datasetCreateSpan != null ? datasetCreate.makeCurrent() : null) {
        ...
```

Ditto to other Span helper methods.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -7492,4 +7541,154 @@ public void testStatementType() throws InterruptedException {
...
+      assertEquals(
+          OTEL_SPAN_IDS_TO_NAMES.get(OTEL_PARENT_SPAN_IDS.get("datasetDelete")),
+          "Test Parent Span");
+      assertEquals(OTEL_PARENT_SPAN_IDS.get("Test Parent Span"), OTEL_PARENT_SPAN_ID);
+      RemoteBigQueryHelper.forceDelete(bigquery, billingModelDataset);
+    }
+  }
+
+  @Test
+  public void TestOpenTelemetryTracingTables() {
```
PhongChuong, line 7619: s/TestOpenTelemetryTracingTables/testOpenTelemetryTracingTables

PhongChuong (commented):
In the case you are not aware. There are 2 additional locations where we perform network calls in TableDataWriteChannel.java and Job.java.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/OpenTelemetryHelper.java
whuffman36: Sorry for the late reply. We already discussed briefly offline about this, but I have updated the code to include this refactor. I still have one function in OpenTelemetryHelper class that several of the `getOtelAttributes()` methods use, but this could be renamed or moved to another file if you think we should get rid of the helper class entirely. Wdyt?

whuffman36:
> In the case you are not aware. There are 2 additional locations where we perform network calls in TableDataWriteChannel.java and Job.java.

Yes, I wanted to keep this PR scoped to just the BigQueryImpl file (though with the refactor that has changed). I was planning to include the calls in TableDataWriteChannel.java and Job.java in a separate PR with a few other methods, such as runWithRetries to make it easier to review and reason about.

shollyman (commented):
My biggest worry with this is the attribute structure here.  This PR appears to define every field of every of every API as it's own attribute, which means we're more likely to introduce conflicts for users trying to filter based on attributes downstream.

This leads me to two questions:
1) should we be putting the full contents of the responses into attributes, vs a more constrained approach (critical fields like identifiers, status, etc)?
2) if we do want to capture the API response, should it just be stringified as a single attribute?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/OpenTelemetryHelper.java
whuffman36: After discussing offline, I have gone ahead and removed the helper class altogether.

whuffman36:
> My biggest worry with this is the attribute structure here. This PR appears to define every field of every of every API as it's own attribute, which means we're more likely to introduce conflicts for users trying to filter based on attributes downstream.
> 
> This leads me to two questions:
> 
> 1. should we be putting the full contents of the responses into attributes, vs a more constrained approach (critical fields like identifiers, status, etc)?
> 2. if we do want to capture the API response, should it just be stringified as a single attribute?

1. I'm of the opinion that since tracing is opt-in and meant to make debugging easier, the more visibility into the data the better. It's better to have visibility into the full contents and not need it than it is to want that visibility and not have it. What do you think?
2. I'm not convinced we should be capturing the response, as it essentially just passed to the user. The user can decide to capture or log the response once received outside of the library code. This design was to capture just the inputs to the API calls. That being said, I think stringifying a response as a single attribute makes the attribute harder to parse and read when debugging, which can defeat the purpose of creating a better debug system.

PhongChuong:
My 2 cents regarding attributes:
Recall that we have 2 layers for the span. The upper level interface (BigQueryImpl.java, Job.java, etc.) and the HttpBigQueryRpc level. I believe we should capture the request and response (not in this PR). Specifically, it should be captured as a request/response* attribute at the HttpBigQueryRpc level. For the interface level (this PR), we should capture any metadata information that would help with debugging any issue that may arise and we should capture it in a structured manner so it is visible. Specifically, attributes in the options are great as they are rarely captured elsewhere. Meanwhile, we can be a little more selective with datatype attributes (TableId vs all the fields in TableInfo).  And if needed, then the request attribute can be used to deep dive.

On a side note, I think we should also clearly state that "Span names and attributes are subject to change without notice" to allow us the flexibility to change the structure as needed.

* For certain responses such as table row data, we probably won't want to capture those values.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -90,6 +95,16 @@ public Builder setUseInt64Timestamps(boolean useInt64Timestamps) {
       return this;
     }
 
+    public Builder setEnableOpenTelemetryTracing(boolean enableOpenTelemetryTracing) {
```
PhongChuong, line 98: Lets mark the builder/getters `@BetaApi` to make it clear that some aspect of this feature is subject to change. Please include a comment for setEnableOpenTelemetryTracing also.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -269,7 +272,18 @@ public Dataset create(DatasetInfo datasetInfo, DatasetOption... options) {
...
             .toPb();
     final Map<BigQueryRpc.Option, ?> optionsMap = optionMap(options);
-    try {
+    Span datasetCreate = null;
+    if (getOptions().isOpenTelemetryTracingEnabled()
+        && getOptions().getOpenTelemetryTracer() != null) {
+      datasetCreate =
+          getOptions()
+              .getOpenTelemetryTracer()
+              .spanBuilder("JAVA_BQ_SDK_datasetCreate")
```
PhongChuong, line 281: Open telemetry has a set of guidelines for span names and attributes. Please see if any are applicable:
[Database](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/#general-database-attributes)
[RPC](https://opentelemetry.io/docs/specs/semconv/attributes-registry/rpc/)
[HTTP](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/)
[client attributes](https://opentelemetry.io/docs/specs/semconv/general/attributes/#client-attributes)

PhongChuong, line 281: For the span name, I think something along the line of:
{package_name}.{methodName} 
I.e., com.google.cloud.bigquery.BigQuery.createDataset
The language can probably be added as an attribute such as:
language: Java
or
library: java-bigquery

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
```diff
@@ -763,4 +764,40 @@ Table toPb() {
   static TableInfo fromPb(Table tablePb) {
     return new BuilderImpl(tablePb).build();
   }
+
+  private static String getFieldAsString(Object field) {
+    return field == null ? "null" : field.toString();
+  }
+
+  public Attributes getOtelAttributes() {
```
PhongChuong, line 772: We probably should change getOtelAttributes to protected and decide again once the feature is more stable.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
whuffman36: Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
whuffman36: I updated the attributes to include a "db.name" field like you mentioned in the email with Lingqing, but I think the client, http, and rpc schemes would be better suited for the HttpBigQueryRpc methods as those are actually doing the network calls.

whuffman36: Good idea, updated the span names to include the package name instead of "JAVA_BQ_SDK_" and added in the language as an attribute as well.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
whuffman36: Done

whuffman36:
I've cleaned up the attributes that we are collecting by removing some fields that are not very useful in debugging. I also added in attributes for language and "db.name" to make the tracing more consistent with python. I've updated the span naming pattern to be more descriptive.

I think it's a good idea to include "Span names and attributes are subject to change without notice", but that seems like it would better to put in the docs rather than the code. What do you guys think?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DatasetId.java
```diff
@@ -84,4 +85,12 @@ DatasetReference toPb() {
   static DatasetId fromPb(DatasetReference datasetRef) {
     return new DatasetId(datasetRef.getProjectId(), datasetRef.getDatasetId());
   }
+
+  protected Attributes getOtelAttributes() {
+    return Attributes.builder()
+        .put("project", this.getProject())
+        .put("dataset", this.getDataset())
+        .put("db.name", this.getDataset())
```
PhongChuong, line 93: From the python side, I think they are using project instead of dataset.
https://github.com/googleapis/python-bigquery/blob/46927479085f13fd326e3f2388f60dfdd37f7f69/google/cloud/bigquery/opentelemetry_tracing.py#L125-L164

ditto elsewhere.

Thinking about it more, adding "db.name" just because python does it seems silly. I vote we remove it. What do you think?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
whuffman36: Done. Thanks for the tip, didn't know you could put strings in the beta annotation.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -691,6 +852,10 @@ public Boolean call() throws IOException {
         return false;
       }
       throw BigQueryException.translateAndThrow(e);
+    } finally {
+      if (routineDelete != null) {
+        routineDelete.end();
+      }
     }
   }
```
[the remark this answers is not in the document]
whuffman36: Good catch! Updated the method to have a span

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1714,6 +2274,53 @@ public com.google.api.services.bigquery.model.TestIamPermissionsResponse call()
...
+        .put("defaultDataset", getFieldAsString(request.getDefaultDataset()))
+        .put("dryRun", getFieldAsString(request.getDryRun()))
+        .put("jobCreationMode", getFieldAsString(request.getJobCreationMode()))
+        .put("jobTimeoutMs", getFieldAsString(request.getJobTimeoutMs()))
+        .put("kind", getFieldAsString(request.getKind()))
+        .put("labels", getFieldAsString(request.getLabels()))
+        .put("location", getFieldAsString(request.getLocation()))
+        .put("maxResults", getFieldAsString(request.getMaxResults()))
+        .put("maximumBytesBilled", getFieldAsString(request.getMaximumBytesBilled()))
+        .put("query", getFieldAsString(request.getQuery()))
```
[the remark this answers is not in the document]
whuffman36: Done, I kept them out of the QueryJobConfiguration Attributes as well.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DatasetId.java
whuffman36: I agree with your sentiment on "db.name", I removed the attribute as it was just a duplicate of "dataset". As for the python version, I'm not sure why they use the project name for "db.name", but I think mapping the attributes of project, dataset, and location make more sense here.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1399,6 +1867,10 @@ public Boolean call() throws IOException {
         return false;
       }
       throw BigQueryException.translateAndThrow(e);
+    } finally {
+      if (jobCancel != null) {
+        jobCancel.end();
+      }
     }
   }
```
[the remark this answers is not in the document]
whuffman36: This is a good point, I have gone ahead and added spans fro the two query functions and listPartitions to make the span chain clearer. It should be more obvious to the user which path the code goes through based on the span chains.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -956,43 +1229,155 @@ public com.google.api.services.bigquery.model.Routine call() throws IOException
...
+    try (Scope routinesListScope = routinesList != null ? routinesList.makeCurrent() : null) {
+      return listRoutines(completeDatasetId, getOptions(), optionMap(options));
+    } finally {
+      if (routinesList != null) {
+        routinesList.end();
+      }
+    }
   }
 
   @Override
```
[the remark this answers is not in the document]
whuffman36: Updated listPartitions to have a span creation block as well.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1219,11 +1624,30 @@ public TableResult listTableData(
   public TableResult listTableData(TableId tableId, Schema schema, TableDataListOption... options) {
```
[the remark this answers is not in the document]
whuffman36: I think ideally, we could have span creation blocks in just the private methods that all of the public overloads eventually lead to. Initially I assumed that I could not go through and remove the static from the private methods, but after discussing offline that could be a viable strategy.

This is definitely something to keep in mind for a future refactor, but it might take a bit more thought.

PhongChuong (dismissed):
Thanks for all the changes.
@shollyman , can you take a look at the attributes to see if we missed anything?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/InsertAllRequest.java
```diff
@@ -479,4 +480,18 @@ public boolean equals(Object obj) {
...
   }
+
+  private static String getFieldAsString(Object field) {
+    return field == null ? "null" : field.toString();
+  }
+
+  public Attributes getOtelAttributes() {
+    return Attributes.builder()
+        .put("table", getFieldAsString(this.getTable().getTable()))
+        .put("row", getFieldAsString(this.getRows()))
```
shollyman, line 491: Capturing where this request is going makes sense (destination table and suffix if present).  Should it be reported the same way as TableId is setup to report?

Row data and the secondary options seems excessive.  Maybe just the number of rows in the request?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobInfo.java
```diff
@@ -375,4 +376,23 @@ public static JobInfo of(JobId jobId, JobConfiguration configuration) {
   static JobInfo fromPb(Job jobPb) {
     return new BuilderImpl(jobPb).build();
   }
+
+  private static String getFieldAsString(Object field) {
+    return field == null ? "null" : field.toString();
+  }
+
+  protected Attributes getOtelAttributes() {
+    return Attributes.builder()
```
shollyman, line 385: of these, jobtype is probably the most valuable (if present) since it allows users to filter workloads.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1714,6 +2344,51 @@ public com.google.api.services.bigquery.model.TestIamPermissionsResponse call()
...
+    for (Option option : options) {
+      attributes =
+          attributes.toBuilder()
+              .put(option.getRpcOption().toString(), option.getValue().toString())
+              .build();
+    }
+    return attributes;
+  }
+
+  private static Attributes otelAttributesFromQueryRequest(QueryRequest request) {
```
shollyman, line 2362: requestId and location are the most identifier-like fields here.

shollyman (commented):
I continue to be wary of replicating many of the API response fields in the span attributes, favoring the "less is more" approach, but I can be convinced otherwise.

whuffman36:
I've gone through and removed most attributes that we were collecting before, keeping mainly just the OP identifiers and time information such as creationTime, lastModifiedTime, etc. Let me know if you think we should shave off any more

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -389,6 +437,18 @@ Job create(JobInfo jobInfo, Supplier<JobId> idProvider, JobOption... options) {
...
     final Map<BigQueryRpc.Option, ?> optionsMap = optionMap(options);
+    Span jobCreate = null;
+    if (getOptions().isOpenTelemetryTracingEnabled()
+        && getOptions().getOpenTelemetryTracer() != null) {
+      jobCreate =
+          getOptions()
+              .getOpenTelemetryTracer()
+              .spanBuilder("com.google.cloud.bigquery.BigQuery.createJob")
+              .setAllAttributes(jobInfo.getJobId().getOtelAttributes())
+              .setAttribute("status", getFieldAsString(jobInfo.getStatus()))
```
shollyman, line 448: status is a submessage.  Should we just report status fields individually?  status.state (pending/running/done), and error presence in the job?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -528,7 +609,24 @@ public Page<Dataset> listDatasets(DatasetListOption... options) {
...
   public Page<Dataset> listDatasets(String projectId, DatasetListOption... options) {
-    return listDatasets(projectId, getOptions(), optionMap(options));
+    Span datasetsList = null;
+    if (getOptions().isOpenTelemetryTracingEnabled()
+        && getOptions().getOpenTelemetryTracer() != null) {
+      datasetsList =
+          getOptions()
+              .getOpenTelemetryTracer()
+              .spanBuilder("com.google.cloud.bigquery.BigQuery.listDatasets")
+              .setAttribute("projectId", projectId)
```
shollyman, line 619: minor:  worth pulling the current page token into an attribute alongside the project for the list RPCs?  It'll be visible in the HTTP request below it as part of the URL.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Schema.java
```diff
@@ -99,4 +100,13 @@ static Schema fromPb(com.google.api.services.bigquery.model.TableSchema tableSch
     }
     return Schema.of(FieldList.fromPb(fields));
   }
+
+  public Attributes getOtelAttributes() {
+    Attributes attributes = Attributes.builder().build();
+    for (Field field : this.getFields()) {
+      attributes =
+          attributes.toBuilder().put("Field: " + field.getName(), field.toString()).build();
```
shollyman, line 108: Do we want BQ schema in the otel attributes at all?

shollyman (commented):
Did some more reading, two other considerations for these attributes (which may require a lot of callsite touches):

* Should the attributes be namespaced so it's clear they're fields from the BQ API?  Right now they're just single term attributes and could conflict with other telemetry.

* I believe otel attribute conventions are snake case rather than camel, so we may want to revisit the multi word attributes.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
whuffman36: Sounds good to me, though I've decided to move those attributes to the Job class instead since during the create() function, job.status() is not yet populated. The field would always be set to null and thus be practically useless for debugging.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
whuffman36: If the page token is supplied as a DatasetListOption, it will already be captured in an attribute here. I agree that adding the page token as an attribute is a good idea, though I think it would be better suited to the http layer rather than the interface.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Schema.java
whuffman36: It is likely overkill, I went ahead and removed the schema from the attributes.

whuffman36:
I renamed all the attributes to be snake case and gave them namespace prefixes to remove any overloading of attributes. I went with the namespace "bq" and created further namespaces depending on the objects, such as "bq.dataset", "bq.table", "bq.query", etc. 

Let me know if you think a different namespace style would be better.

PhongChuong (approved):

shollyman (approved):
Thanks for slogging through all this!

PR #3768, opened 2025-04-30
feat(bigquery): Add support for reservation field in jobs.

whuffman36:
This PR adds support to assign the reservation field in JobConfiguration when creating a job. This field is also propagated to QueryRequestInfo.


shollyman (approved):
Thanks for pulling this together so quickly.

PR #3775, opened 2025-05-05
fix: fix graalvm native test B

diegomarquezp:
WIP

### DIFF
--- a/.kokoro/presubmit/integration.cfg
+++ b/.kokoro/presubmit/integration.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_b:3.46.2"
 }
 
 env_vars: {
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -60,14 +60,8 @@ javadoc)
     RETURN_CODE=$?
     ;;
 integration)
-    mvn -B ${INTEGRATION_TEST_ARGS} \
-      -ntp \
-      -Penable-integration-tests \
-      -DtrimStackTrace=false \
-      -Dclirr.skip=true \
-      -Denforcer.skip=true \
-      -fae \
-      verify
+    # Run Unit and Integration Tests with Native Image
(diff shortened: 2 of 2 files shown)


google-cloud-bigquery/pom.xml
```diff
@@ -157,6 +157,12 @@
     </dependency>
 
     <!-- Test dependencies -->
+	<dependency>
```
PhongChuong, line 160: extra spacing

.kokoro/presubmit/integration.cfg
```diff
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_b:3.46.2"
```
PhongChuong, line 6: It feels weird that we would switch all of integration tests to use this graalvm image. I think we should consider:
- if the graalvm does not need to be part of presubmit integration tests, then we should remove it from the test suite.
- if the graalvm does need to be part of presubmit tests, then we should create its own presubmit config instead of overriding the integration test default image.

google-cloud-bigquery/pom.xml
diegomarquezp: Done.

.kokoro/presubmit/integration.cfg
diegomarquezp: This was only done to confirm the configuration works in the continuous graalvm tests. Unfortunately, the graalvm tests only run after merging the PRs, so I temporarily overrode the IT to use this image instead so we can confirm it works in the presubmits.

Do you think we should also add a presubmit for GraalVM tests to avoid this kind of configuration testing?

PhongChuong: Lets add the presubmit tests for GraalVMs. If the tests takes too long or is flaky, then we can revisit removing it from presubmit.
Thank you for looking into this.

diegomarquezp:
@PhongChuong the jobs were successfully rolled out as presubmits.

image

.kokoro/presubmit/integration.cfg
diegomarquezp: Done. Left a comment in main thread showing the successfully added presubmits. Thanks for the feedback.

PR #3803, opened 2025-05-28
feat(bigquery): Support Fine Grained ACLs for Datasets

whuffman36:
This PR adds the options dataset_view and update_mode to the get and patch dataset requests respectively. These new options decide whether the metadata and or ACL information should be checked when retrieving or updating a dataset.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
@@ -77,6 +77,40 @@ public interface BigQuery extends Service<BigQueryOptions> {
     }
   }
 
+  enum DatasetView {
+    FULL("full"),
+    METADATA("metadata"),
+    ACL("acl");
+
+    private final String view;
+
+    DatasetView(String view) {
+      this.view = view;
+    }
+
+    @Override
+    public String toString() {
+      return view.toUpperCase();
+    }
+  }
+
+  enum DatasetUpdateMode {
+    UPDATE_FULL("update_full"),
+    UPDATE_METADATA("update_metadata"),
+    UPDATE_ACL("update_acl");
+
+    private final String updateMode;
+
+    DatasetUpdateMode(String updateMode) {
+      this.updateMode = updateMode;
+    }
+
+    @Override
+    public String toString() {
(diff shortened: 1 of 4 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -77,6 +77,40 @@ public String getSelector() {
     }
   }
 
+  enum DatasetView {
+    FULL("full"),
```
shollyman, line 81: Can we define these using the casing the service describes (e.g. uppercase), rather than performing case manipulation in the toString()?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -77,6 +77,40 @@ public String getSelector() {
...
+    }
+
+    @Override
+    public String toString() {
+      return view.toUpperCase();
+    }
+  }
+
+  enum DatasetUpdateMode {
+    UPDATE_FULL("update_full"),
```
shollyman, line 98: same here re:case handling

shollyman (dismissed):
Minor comments about the enum definitions, but LGTM otherwise.  Thanks for putting this together!

shollyman (dismissed):
one other minor thing: do we want to document the default _UNSPECIFIED enum values?  I'm fine either way.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
whuffman36: Fixed!

whuffman36:
The _UNSPECIFIED enum value seems a bit redundant to me, since the proto definition says that it will just default to FULL if unspecified. What do the other languages do here?

shollyman:
> The _UNSPECIFIED enum value seems a bit redundant to me, since the proto definition says that it will just default to FULL if unspecified. What do the other languages do here?

This is a manifestation of https://google.aip.dev/126 which specifies the default (0 value) enum handles the unspecified case.  It's a way of explicitly stating "use the service's default" value, which is documented as meaning "full" today but could potentially change in the future.

PR #3842, opened 2025-06-10
feat(bigquery): Integrate Otel into retries, jobs, and more

whuffman36:
The more here refers to `TableDataWriteChannel`. This PR also changes the signature of `BigQueryRetryHelper.runWithRetries()` to include a `boolean` and Otel `Tracer` as parameters.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
@@ -476,7 +476,9 @@ class ConnectionImpl implements Connection {
               bigQueryOptions.getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
               bigQueryOptions.getClock(),
-              retryConfig);
+              retryConfig,
+              bigQueryOptions.isOpenTelemetryTracingEnabled(),
+              bigQueryOptions.getOpenTelemetryTracer());
     } catch (BigQueryRetryHelperException e) {
       throw BigQueryException.translateAndThrow(e);
     }
@@ -932,7 +934,9 @@ class ConnectionImpl implements Connection {
               bigQueryOptions.getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
               bigQueryOptions.getClock(),
-              EMPTY_RETRY_CONFIG);
+              EMPTY_RETRY_CONFIG,
+              bigQueryOptions.isOpenTelemetryTracingEnabled(),
+              bigQueryOptions.getOpenTelemetryTracer());
     } catch (BigQueryRetryHelperException e) {
       if (e.getCause() instanceof BigQueryException) {
         if (((BigQueryException) e.getCause()).getCode() == HTTP_NOT_FOUND) {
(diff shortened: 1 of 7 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -329,47 +360,64 @@ public TableResult getQueryResults(QueryResultsOption... options)
...
-    TableId table =
-        ((QueryJobConfiguration) getConfiguration()).getDestinationTable() == null
-            ? ((QueryJobConfiguration) job.getConfiguration()).getDestinationTable()
-            : ((QueryJobConfiguration) getConfiguration()).getDestinationTable();
-    TableResult tableResult =
-        bigquery.listTableData(
-            table, response.getSchema(), listOptions.toArray(new TableDataListOption[0]));
-    TableResult tableResultWithJobId = tableResult.toBuilder().setJobId(job.getJobId()).build();
-    return tableResultWithJobId;
+    Span getQueryResults = null;
```
PhongChuong, line 363: I wonder if it's better to immediately start the span after checkNotDryRun("getQueryResults") even though no network calls are being performed until this section. I feel like it does make the code a little more consistent and in the unlikely case that network calls are added inadvertently to it in the future.

Note: I don't feel particularly strongly one way or the other but it's something to consider.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
```diff
@@ -476,7 +476,9 @@ private BigQueryResult queryRpc(
               bigQueryOptions.getRetrySettings(),
               BigQueryBaseService.BIGQUERY_EXCEPTION_HANDLER,
               bigQueryOptions.getClock(),
-              retryConfig);
+              retryConfig,
```
PhongChuong, line 479: Since we're not supporting open telemetry in the Connection interface. I think it might be better to pass in (false, null) instead of (bigQueryOptions.isOpenTelemetryTracingEnabled(), and bigQueryOptions.getOpenTelemetryTracer()), respectively, since the later would create spans in the child calls without the encapsulating parent span in ConnectionImpl.

We probably should also explicitly call out that open telemetry is not supported in the Connection Interface in the further sample + README.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -463,14 +554,31 @@ public boolean shouldRetry(Throwable prevThrowable, Job prevResponse) {
...
     checkNotDryRun("reload");
-    Job job = bigquery.getJob(getJobId(), options);
-    if (job != null && job.getStatus().getError() != null) {
-      throw new BigQueryException(
-          job.getStatus().getExecutionErrors() == null
-              ? ImmutableList.of(job.getStatus().getError())
-              : ImmutableList.copyOf(job.getStatus().getExecutionErrors()));
-    }
-    return job;
+    Span reload = null;
```
PhongChuong, line 557: Should we add a parent span for either:
public Job waitFor
or
private Job waitForInternal
?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ConnectionImpl.java
PhongChuong: I think stating it in:
1. Root README
2. As a comment in the open telemetry sample.
Would be sufficient.

PhongChuong (approved):

shollyman (approved):

PR #3995, opened 2025-11-05
feat: new queryWithTimeout method for customer-side wait

logachev:
Existing `BigQuery.query()` method always returns TableResult for both long-running a jobless queries. 
One of the issues with this approach is that for long-running operations, consumer app has no way to track the Job because JobId is unknown until after query is complete. It prevents consumer apps from cancelling queries if they're running too long or if customer wants to cancel jobs due various reasons.

This PR introduces new `queryWithTimeout` method that returns either `TableResult` object for short queries or `Job` object for long-running query. 
It will allow to manipulate `Job` before query is finished (can be relevant for queries that take more than a few minutes to complete).

It also refactors `query` methods to leverage new `queryNoWait` method to avoid code duplication which fixed an issue that `query()` version with `JobId` was never leveraging jobless query feature.

Added integration & unit test for new method covering timeout feature.

### DIFF
--- a/google-cloud-bigquery/clirr-ignored-differences.xml
+++ b/google-cloud-bigquery/clirr-ignored-differences.xml
@@ -119,6 +119,11 @@
   </difference>
   <difference>
     <differenceType>7012</differenceType>
+    <className>com/google/cloud/bigquery/BigQuery</className>
+    <method>java.lang.Object queryWithTimeout(com.google.cloud.bigquery.QueryJobConfiguration, com.google.cloud.bigquery.JobId, java.lang.Long, com.google.cloud.bigquery.BigQuery$JobOption[])</method>
+  </difference>
+  <difference>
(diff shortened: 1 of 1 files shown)


jinseopkim0 (changes requested):
Thank you for this PR. The clirr CI is failing with the following error:
```
Error:  7012: com.google.cloud.bigquery.BigQuery: Method 'public java.lang.Object queryWithTimeout(com.google.cloud.bigquery.QueryJobConfiguration, com.google.cloud.bigquery.JobId, java.lang.Long, com.google.cloud.bigquery.BigQuery$JobOption[])' has been added to an interface
```

Could you please add a default implementation to the queryWithTimeout method? An empty implementation throwing an exception would suffice.

Also, would it make sense to add `@InternalApi` annotation to this method?

jinseopkim0 (approved):

PR #4010, opened 2025-12-03
feat: Introduce DataFormatOptions to configure the output of BigQuery data types

lqiu96:
See b/447623336 for more information. 

Introduces a DataFormatOptions class to configure how to format the data for BigQuery outputs. It includes the existing functionality for `useInt64Timestamps` that is currently a setting in BigQueryOptions. Those settings are absorbed into the new DataFormatOptions class. 

This PR does not change the logic for the existing BigQuery code, but converts any existing use cases of BigQueryOptions's `useInt64Timestamps` to now reference BigQueryOption's DataFormatOptions.useInt64Timestamps value instead.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DataFormatOptions.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DataFormatOptions.java
@@ -70,17 +70,4 @@ public abstract class DataFormatOptions implements Serializable {
     request.setTimestampOutputFormat(timestampFormatOptions().toString());
     return request;
   }
-
-  static DataFormatOptions fromPb(
-      com.google.api.services.bigquery.model.DataFormatOptions request) {
-    DataFormatOptions.Builder builder = newBuilder();
-    if (request.getUseInt64Timestamp() != null) {
-      builder.useInt64Timestamp(request.getUseInt64Timestamp());
-    }
-    if (request.getTimestampOutputFormat() != null) {
-      builder.timestampFormatOptions(
-          TimestampFormatOptions.valueOf(request.getTimestampOutputFormat()));
-    }
-    return builder.build();
-  }
 }


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -94,11 +98,32 @@ public Builder setLocation(String location) {
...
+    /**
+     * This setter is marked as Obsolete. Prefer {@link #setDataFormatOptions(DataFormatOptions)} to
+     * set the int64timestamp configuration instead.
+     *
+     * <p>If useInt64Timestamps value is set in here and via DataFormatOptions, the
+     * DataFormatOptions configuration value is used.
+     *
+     * <p>{@code DataFormatOptions.newBuilder().setUseInt64Timestamp(...).build()}
+     */
+    @ObsoleteApi("Use setDataFormatOptions(DataFormatOptions) instead")
```
jinseopkim0: Is there a reason you would prefer `@ObsoleteApi` over `@Deprecated`?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/DataFormatOptions.java
```diff
@@ -0,0 +1,86 @@
...
+
+  static DataFormatOptions fromPb(
+      com.google.api.services.bigquery.model.DataFormatOptions request) {
+    DataFormatOptions.Builder builder = newBuilder();
+    if (request.getUseInt64Timestamp() != null) {
+      builder.useInt64Timestamp(request.getUseInt64Timestamp());
+    }
+    if (request.getTimestampOutputFormat() != null) {
+      builder.timestampFormatOptions(
+          TimestampFormatOptions.valueOf(request.getTimestampOutputFormat()));
```
jinseopkim0: If BQ backend adds a new format, this would fail, right? Can we try-catch, so that we would fall back to TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED?

jinseopkim0 (changes requested):

jinseopkim0 (approved):

PR #4028, opened 2025-12-12
chore: Use assertThrows in tests

lqiu96:


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -2808,15 +2819,15 @@ public void testUpdateNonExistingTable() {
...
-    }
+    BigQueryException exception =
+        assertThrows(
+            "BigQueryException was expected",
+            BigQueryException.class,
+            () -> bigquery.update(tableInfo));
+    BigQueryError error = exception.getError();
+    assertNotNull(error);
+    assertEquals("notFound", error.getReason());
+    assertNotNull(error.getMessage());
```
mpeddada1: Perhaps we could also simplify the remaining assert statements further by using the Truth.assertThat?
Example: assertThat(exception).hasMessageThat().contains("...");

PR #4066, opened 2026-01-16
chore(jdbc): JDBC presubmit & nightly pipelines

logachev:
- Adding JDBC to the parent pom.xml
- Excluding JDBC project from integration/graal-vm Kokoro runs
- Add 2 new cloudbuilds: presubmit & nightly
- Update versions.txt to contain jdbc version

Note: for now I also enabled ignoring "used but undeclared dependencies". I'm not sure what's the right path given that those dependencies are entirely due to BQ/BQ-Storage dependencies. Not sure if duplicating the list makes a lot of sense

### DIFF
--- a/google-cloud-bigquery-jdbc/pom.xml
+++ b/google-cloud-bigquery-jdbc/pom.xml
@@ -103,110 +103,90 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>api-common</artifactId>
-      <version>2.56.1</version>
     </dependency>
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax</artifactId>
-      <version>2.73.1</version>
     </dependency>
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax-grpc</artifactId>
-      <version>2.73.1</version>
     </dependency>
     <dependency>
       <groupId>com.google.api.grpc</groupId>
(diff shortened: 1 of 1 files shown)


versions.txt
```diff
@@ -1,4 +1,5 @@
 # Format:
 # module:released-version:current-version
 
-google-cloud-bigquery:2.57.2:2.57.2
\ No newline at end of file
+google-cloud-bigquery:2.57.2:2.57.2
+google-cloud-bigquery-jdbc:0.1.0:0.1.1-SNAPSHOT
```
lqiu96: I believe this will see the current version of v0.1.1-SNAPSHOT and releasing it will release it as v0.2.0. Just want to confirm if this is ok with you.

lqiu96:
> Note: for now I also enabled ignoring "used but undeclared dependencies". I'm not sure what's the right path given that those dependencies are entirely due to BQ/BQ-Storage dependencies. Not sure if duplicating the list makes a lot of sense

I feel that we should duplicate in the module instead of having it rely on the transitive version from BQ + BQStorage. I think that if a module uses a dep, it should declare the dep. Ideally, the versions should be managed by shared-deps.

We have users who don't always use the libraries-bom and we don't promise compatibility for those cases. However,  declaring the versions in the module helps minimize the chances of these incompatibilities (users have to override each dep in jdbc instead of just overriding BQ + BQStorage).

PR #4076, opened 2026-01-21
chore(jdbc): add nightly JAR build

logachev:
Adding a new step in Nightly validation to build Uber JAR/Shaded Uber JAR & ZIP archive with Thin jar & upload to internal GCS bucket. 

Updated pom.xml in a way that `mvn package` builds both - thin & uber JARs. Running `mvn deploy` copies both of them to the repo, so next time release is cut I'd expect it to go to maven too.

### DIFF
--- a/google-cloud-bigquery-jdbc/pom.xml
+++ b/google-cloud-bigquery-jdbc/pom.xml
@@ -96,21 +96,22 @@
                   <relocations>
                     <relocation>
                       <pattern>com</pattern>
-                      <shadedPattern>shaded.bqjdbc.com</shadedPattern>
+                      <shadedPattern>com.google.bqjdbc.shaded.com</shadedPattern>
                       <excludes>
                         <exclude>com.google.cloud.bigquery.*</exclude>
+                        <exclude>com.google.cloud.bigquery.jdbc.*</exclude>
                       </excludes>
                     </relocation>
                     <relocation>
                       <pattern>org</pattern>
-                      <shadedPattern>shaded.bqjdbc.org</shadedPattern>
+                      <shadedPattern>com.google.bqjdbc.shaded.org</shadedPattern>
                       <excludes>
                         <exclude>org.conscrypt.*</exclude>
                       </excludes>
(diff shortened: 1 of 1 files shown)


.kokoro/jdbc-release.sh
```diff
@@ -0,0 +1,56 @@
...
+# Thin release
+mkdir -p "./${FOLDER}"
+make docker-package PACKAGE_DESTINATION="$(pwd)/${FOLDER}"
+ZIP_FILE=$(find . -wholename "./${FOLDER}/*zip" -print -quit)
+
+gsutil cp "${ZIP_FILE}" "${NIGHTLY_BUILD_DESTINATION}/${JAR_NAME}.zip"
+rm -rf "${FOLDER}"
+
+# Update latest version
+gsutil cp "${NIGHTLY_BUILD_DESTINATION}/${JAR_NAME}.zip" "${BUCKET}/google-cloud-bigquery-jdbc-latest.zip"
```
lqiu96: qq, how does this script do auth to determine if you have permission to upload?

.kokoro/jdbc-release.sh
```diff
@@ -0,0 +1,56 @@
...
+set -euo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "${DIR}/../google-cloud-bigquery-jdbc"
+
+FOLDER=release
+DATE=$(date '+%Y-%m-%d')
+COMMIT=$(git rev-parse --short HEAD)
+VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+BUCKET=gs://bq_devtools_release_private/drivers/jdbc
```
lqiu96: I'm guessing customers already pull from this bucket and it's public. Just want to check if this being public is fine/ should we put this as an env var?

logachev: That's internal bucket; I moved it to CloudBuild env

PR #4086, opened 2026-01-28
fix(jdbc): return null for invalid URLs in `connect()` to comply with…

keshavdandeva:
b/479264130

According to the `java.sql.Driver.connect()` specification, drivers should return `null` instead of throwing an exception when they cannot accept a given URL. This allows DriverManager to iterate through all registered drivers until it finds one that can handle the URL.


lqiu96:
Changes LGTM. Going to rescind the approval to test that approval permissions for jdbc-team.

PR #4089, opened 2026-02-02
feat(jdbc): add `HttpTransportOptions` timeout configuration

keshavdandeva:
b/454382967

Adds support for configuring `HttpConnectTimeout` and `HttpReadTimeout` in the BigQuery JDBC driver via connection properties or the JDBC URL. This allows users to control the connect and read timeouts for the underlying HTTP transport


google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcProxyUtility.java
```diff
@@ -117,18 +118,39 @@ static HttpTransportOptions getHttpTransportOptions(
...
+
+    HttpTransportOptions.Builder httpTransportOptionsBuilder = HttpTransportOptions.newBuilder();
+    if (hasProxyOrSsl) {
+      httpTransportOptionsBuilder.setHttpTransportFactory(
+          getHttpTransportFactory(
+              proxyProperties, sslTrustStorePath, sslTrustStorePassword, callerClassName));
+    }
+
+    if (hasTimeoutConfig) {
+      httpTransportOptionsBuilder.setConnectTimeout(
```
logachev: I'd change this to set property only if it is provided, otherwise allow it to use httpTransport default value rather than overriding it

logachev (approved):

PR #4100, opened 2026-02-11
feat(jdbc): add retry mechanism for Arrow and JSON result streams

keshavdandeva:
b/473629793

### DIFF
--- a/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
+++ b/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
@@ -56,6 +56,7 @@ import com.google.cloud.bigquery.storage.v1.ReadSession;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
+import com.google.common.util.concurrent.Uninterruptibles;
 import java.lang.ref.ReferenceQueue;
 import java.sql.Connection;
 import java.sql.ResultSet;
@@ -814,9 +815,6 @@ public class BigQueryStatement extends BigQueryNoOpsStatement {
           int retryCount = 0;
           try {
             // Use the first stream to perform reading.
-            if (readSession.getStreamsCount() == 0) {
-              return;
-            }
             String streamName = readSession.getStreams(0).getName();
 
             while (true) {
@@ -836,7 +834,8 @@ public class BigQueryStatement extends BigQueryNoOpsStatement {
                   }
 
                   ArrowRecordBatch currentBatch = response.getArrowRecordBatch();
-                  arrowBatchWrapperBlockingQueue.put(BigQueryArrowBatchWrapper.of(currentBatch));
+                  Uninterruptibles.putUninterruptibly(
+                      arrowBatchWrapperBlockingQueue, BigQueryArrowBatchWrapper.of(currentBatch));
                   rowsRead += response.getRowCount();
                 }
                 break;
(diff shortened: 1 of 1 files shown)


google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -992,39 +1033,64 @@ Thread runNextPageTaskAsync(
...
+                startTimeLoop = System.currentTimeMillis();
+                LOG.fine(
+                    "Fetched %d results from the server in %d ms.",
+                    querySettings.getMaxResultPerPage(),
+                    (int) ((System.nanoTime() - startTime) / 1000000));
+              } catch (com.google.cloud.bigquery.BigQueryException ex) {
+                if (ex.getCode() == 404) {
+                  throw ex;
+                }
+                long elapsedSecs = (System.currentTimeMillis() - startTimeLoop) / 1000;
```
logachev: Better to use `System.nanotime()` to measure duration in general (`currentTimeMillis()` can go backwards while `nanotime` cannot)

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -808,30 +810,77 @@ Thread populateArrowBufferedQueue(
 
     Runnable arrowStreamProcessor =
         () -> {
+          long rowsRead = 0;
+          int retryCount = 0;
           try {
             // Use the first stream to perform reading.
+            if (readSession.getStreamsCount() == 0) {
```
logachev: I think this check is wrong. We either should ensure we have single stream & raise exception (or it will act like "no more rows to read", right?) otherwise or just remove this check. Next line will raise exception is `getStreamCount()` is 0.

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -808,30 +810,77 @@ Thread populateArrowBufferedQueue(
...
+                  try {
+                    arrowBatchWrapperBlockingQueue.put(
+                        BigQueryArrowBatchWrapper.ofError(new BigQueryJdbcRuntimeException(e)));
+                  } catch (InterruptedException ie) {
+                    Thread.currentThread().interrupt();
+                  }
+                  break;
+                }
+                retryCount++;
+                LOG.info(
```
logachev: log.warning

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -992,45 +1045,76 @@ Thread runNextPageTaskAsync(
     // This thread makes the RPC calls and paginates
     Runnable nextPageTask =
         () -> {
-          // results.getPageToken();
-          String pageToken = firstPageToken;
+          int retryCount = 0;
```
logachev: Not sure if retries for paginated API is needed. SDK layer will do a few retries for us.. if we want more retries I assume SDK has parameters to configure it. Can you double-check?

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -808,30 +810,94 @@ Thread populateArrowBufferedQueue(
...
+                }
                 break;
+              } catch (com.google.api.gax.rpc.ApiException e) {
+                if (e.getStatusCode().getCode()
+                    == com.google.api.gax.rpc.StatusCode.Code.NOT_FOUND) {
+                  LOG.warning("Read session expired or not found: %s", e.getMessage());
+                  try {
+                    arrowBatchWrapperBlockingQueue.put(
+                        BigQueryArrowBatchWrapper.ofError(new BigQueryJdbcRuntimeException(e)));
+                  } catch (InterruptedException ie) {
```
logachev: If you catch Interrupted exception, thread was already interrupted, so there is no need in this try/catch block. Can just leave `put()`

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -808,30 +810,94 @@ Thread populateArrowBufferedQueue(
...
             }
 
-          } catch (RuntimeException | InterruptedException e) {
+          } catch (InterruptedException e) {
             LOG.log(
                 Level.WARNING,
                 "\n" + Thread.currentThread().getName() + " Interrupted @ arrowStreamProcessor",
                 e);
+            try {
+              arrowBatchWrapperBlockingQueue.put(
```
logachev: That also looks like `put(ofError)` but with a lot of complications.

logachev (approved):

PR #4107, opened 2026-02-17
feat(jdbc): enforce strict JDBC URL parsing and sync `DataSource` properties

keshavdandeva:
b/429272203

This PR refactors the JDBC connection string parsing logic to be stricter and more robust, preventing silent failures due to typos. It also synchronizes `DataSource` properties with `BigQueryConnection` to ensure full feature parity.

### DIFF
--- a/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
+++ b/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
@@ -665,9 +665,6 @@ final class BigQueryJdbcUrlUtility {
    * @throws BigQueryJdbcRuntimeException if an unknown property is found or the URL is malformed.
    */
   static Map<String, String> parseUrl(String url) {
-    if (url == null) {
-      return Collections.emptyMap();
-    }
     return PARSE_CACHE.computeIfAbsent(url, BigQueryJdbcUrlUtility::parseUrlInternal);
   }
 
--- a/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/DataSource.java
+++ b/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/DataSource.java
@@ -17,6 +17,8 @@
 package com.google.cloud.bigquery.jdbc;
 
 import com.google.cloud.bigquery.exception.BigQueryJdbcException;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableMap;
 import java.io.PrintWriter;
 import java.sql.Connection;
 import java.sql.DriverManager;
@@ -351,14 +353,7 @@ public class DataSource implements javax.sql.DataSource {
     if (map == null || map.isEmpty()) {
       return "";
     }
-    StringBuilder sb = new StringBuilder();
-    for (Map.Entry<String, String> entry : map.entrySet()) {
-      if (sb.length() > 0) {
-        sb.append(",");
-      }
-      sb.append(entry.getKey()).append("=").append(entry.getValue());
-    }
-    return sb.toString();
+    return Joiner.on(",").withKeyValueSeparator("=").join(map);
   }
 
   @Override
@@ -451,7 +446,7 @@ public class DataSource implements javax.sql.DataSource {
   }
 
   public void setQueryProperties(Map<String, String> queryProperties) {
-    this.queryProperties = queryProperties;
+    this.queryProperties = queryProperties == null ? null : ImmutableMap.copyOf(queryProperties);
   }
 
   public Map<String, String> getQueryProperties() {
@@ -711,7 +706,7 @@ public class DataSource implements javax.sql.DataSource {
   }
 
   public void setLabels(Map<String, String> labels) {
-    this.labels = labels;
+    this.labels = labels == null ? null : ImmutableMap.copyOf(labels);
   }
 
   public String getRequestReason() {


google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -591,6 +597,32 @@ final class BigQueryJdbcUrlUtility {
...
+      map.put(p.getName().toUpperCase(), p.getName());
+    }
+    for (String p : OVERRIDE_PROPERTIES) {
+      map.put(p.toUpperCase(), p);
+    }
+    for (String p : BYOID_PROPERTIES) {
+      map.put(p.toUpperCase(), p);
+    }
+    map.put(PARTNER_TOKEN_PROPERTY_NAME.toUpperCase(), PARTNER_TOKEN_PROPERTY_NAME);
+    map.put(ENDPOINT_OVERRIDES_PROPERTY_NAME.toUpperCase(), ENDPOINT_OVERRIDES_PROPERTY_NAME);
```
logachev: I'd suggest adding all these properties to some property groups rather than maintain a list here. (NETWORK_PROPERTIES or just generic VALID_PROPERTIES)

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -601,12 +633,66 @@ private BigQueryJdbcUrlUtility() {}
    * @return The String value of the property, or the default value if the property is not found.
    */
   static String parseUriProperty(String uri, String property) {
-    Pattern pattern = Pattern.compile(String.format("(?is)(?:;|\\?)%s=(.*?)(?:;|$)", property));
-    Matcher matcher = pattern.matcher(uri);
-    if (matcher.find() && matcher.groupCount() == 1) {
-      return CharEscapers.decodeUriPath(matcher.group(1));
+    Map<String, String> map = parseUrl(uri);
```
logachev: Need to cache results so we're not parsing it for every single property

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -601,12 +633,66 @@ private BigQueryJdbcUrlUtility() {}
...
+    if (url == null) {
+      return map;
+    }
+
+    int start = url.indexOf(';');
+    if (start == -1) {
+      return map;
+    }
+
+    String urlToParse = url.substring(start + 1);
```
logachev: nit: can just split the string & ignore first value given it is a URL prefix

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -601,12 +633,66 @@ private BigQueryJdbcUrlUtility() {}
...
+
+    // Parse PartnerToken separately as it contains ';'
+    StringBuilder urlBuilder = new StringBuilder(urlToParse);
```
logachev: Why using StringBuilder?

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -601,12 +633,66 @@ private BigQueryJdbcUrlUtility() {}
...
+    }
+
+    String[] parts = urlToParse.split(";");
+    for (String part : parts) {
+      if (part.trim().isEmpty()) {
+        continue;
+      }
+      String[] kv = part.split("=", 2);
+      String key = kv[0].trim();
+      if (kv.length == 1) {
```
logachev: That seems easier to read? 
```
String key = kv[0].trim().toUpperCase();
if (kv.length != 2 || !PROPERTY_NAME_MAP.containsKey(key) ) {
  throw new BigQueryJdbcRuntimeException(String.format("Wrong value or unknown setting found in connection string: %s", part.substring(0, Math.min(part.length(), 12)))
}

String value = kv[1].trim();
map.put(key, CharEscapers.decodeUriPath(value));
```

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -622,7 +708,14 @@ static String appendPropertiesToURL(String url, String callerClassName, Properti
     for (Entry<Object, Object> entry : properties.entrySet()) {
       if (entry.getValue() != null && !"".equals(entry.getValue())) {
         LOG.finest("Appending %s with value %s to URL", entry.getKey(), entry.getValue());
-        urlBuilder.append(";").append(entry.getKey()).append("=").append(entry.getValue());
+        try {
+          String encodedValue =
+              URLEncoder.encode((String) entry.getValue(), StandardCharsets.UTF_8.name())
```
logachev: We already have `guava` dependency, so can just use it instead: https://guava.dev/releases/23.0/api/docs/com/google/common/net/UrlEscapers.html

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -591,6 +597,32 @@ final class BigQueryJdbcUrlUtility {
                               + " header.")
                       .build())));
 
+  private static final Map<String, String> PROPERTY_NAME_MAP;
+
+  static {
+    Map<String, String> map = new HashMap<>();
+    for (BigQueryConnectionProperty p : VALID_PROPERTIES) {
+      map.put(p.getName().toUpperCase(), p.getName());
```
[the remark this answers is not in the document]
keshavdandeva: No, this map is for normalizing propertry names (mapping `UPPER_CASE` -> CamelCase) to support case-insensitive property keys in the connection URL. 
Default values are handled separately during property retrieval.

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/DataSource.java
```diff
@@ -78,6 +78,17 @@ public class DataSource implements javax.sql.DataSource {
   private Integer metadataFetchThreadCount;
   private String sslTrustStorePath;
   private String sslTrustStorePassword;
+  private String labels;
```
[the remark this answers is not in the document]
keshavdandeva: Good catch! Also, I realised for properties like `labels` and `queryProperties` we needed to serialize maps

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -601,12 +633,66 @@ private BigQueryJdbcUrlUtility() {}
    * @return The String value of the property, or the default value if the property is not found.
    */
   static String parseUriProperty(String uri, String property) {
-    Pattern pattern = Pattern.compile(String.format("(?is)(?:;|\\?)%s=(.*?)(?:;|$)", property));
-    Matcher matcher = pattern.matcher(uri);
-    if (matcher.find() && matcher.groupCount() == 1) {
-      return CharEscapers.decodeUriPath(matcher.group(1));
+    Map<String, String> map = parseUrl(uri);
+    if (PROPERTY_NAME_MAP.containsKey(property.toUpperCase())) {
```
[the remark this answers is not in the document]
logachev: hm, ok. No need to change it in this PR, but I think we should get rid of `parseStringProperty` handling default values. We already have defaultValue() available in various property maps (e.g.  [link](https://github.com/googleapis/java-bigquery/blob/main/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java#L208)) 

I don't see a lot of benefit in using map to map upper-to-camel case.. We can just normalize keys and always use upper case instead.

PR #4116, opened 2026-02-25
feat(jdbc): centralize connection properties from `BigQueryConnection` and `DataSource`

keshavdandeva:
b/484306563

### DIFF
--- a/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
+++ b/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
@@ -935,7 +935,7 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
       bigQueryOptions.setTransportOptions(this.httpTransportOptions);
     }
 
-    BigQueryOptions options = bigQueryOptions.setHeaderProvider(HEADER_PROVIDER).build();
+    BigQueryOptions options = bigQueryOptions.setHeaderProvider(this.headerProvider).build();
     options.setDefaultJobCreationMode(
         this.useStatelessQueryMode
             ? JobCreationMode.JOB_CREATION_OPTIONAL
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -125,7 +125,7 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
   int writeAPIActivationRowCount;
   int writeAPIAppendRowCount;
   int requestGoogleDriveScope;
-  List<String> additionalProjects;
+  String additionalProjects;
```
logachev: Why changing the type?

keshavdandeva: Because I realized the User Guide specifies a comma-separated string, we are keeping the property as `String` in both DataSource and BigQueryConnection to follow the spec and avoid API breakage or problems with any current tools that might switch to our driver. 
Parsing happens on-demand in `BigQueryDatabaseMetaData` where the list is actually needed.

google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -199,94 +184,27 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
...
-    this.destinationDatasetExpirationTime =
-        BigQueryJdbcUrlUtility.parseLongProperty(
-            url,
-            BigQueryJdbcUrlUtility.DESTINATION_DATASET_EXPIRATION_TIME_PROPERTY_NAME,
-            BigQueryJdbcUrlUtility.DEFAULT_DESTINATION_DATASET_EXPIRATION_TIME_VALUE,
-            this.connectionClassName);
-    this.kmsKeyName =
-        BigQueryJdbcUrlUtility.parseStringProperty(
-            url, BigQueryJdbcUrlUtility.KMS_KEY_NAME_PROPERTY_NAME, null, this.connectionClassName);
+    this.location = ds.getLocation();
```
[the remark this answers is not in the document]
keshavdandeva: Yeah we can do that. It will be better than this but I would suggest to do it in a separate PR so it is easier to review and test. Also, might need a bit checking if it wont cause any issues

PR #4126, opened 2026-03-02
feat: add initial opentelemetry tracing to big query HTTP requests

ldetmer:
This feature adds the ability to enable open telemetry tracing on all HTTP requests.  

It reuses the existing client options setting [enableOpenTelemetryTracing](https://github.com/googleapis/java-bigquery/blob/0b0c1ceb309a19d51f385264f2d4722b8b6339d7/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java#L77) to enable it, and then wraps the existing [HttpRequestInitializer](https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/http/HttpRequestInitializer.java) to intercept and add tracing.

This PR only contains the initial basic general/http attributes.  Separate PRs will contain additional attributes.  

Tested via sample test program and validated attributes show up in cloud trace:
https://screenshot.googleplex.com/AQJp4Nbb6oVbgAk

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
@@ -109,7 +109,7 @@ public class HttpBigQueryRpc implements BigQueryRpc {
     HttpRequestInitializer initializer = transportOptions.getHttpRequestInitializer(options);
 
     String resolvedBigQueryRootUrl = options.getResolvedApiaryHost("bigquery");
-    // Wrap with tracing initializer if OpenTelemetry is enabled
+
     if (options.isOpenTelemetryTracingEnabled() && options.getOpenTelemetryTracer() != null) {
       initializer =
           new HttpTracingRequestInitializer(
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
@@ -55,10 +55,20 @@ public class HttpTracingRequestInitializer implements HttpRequestInitializer {
       AttributeKey.longKey("http.response.body.size");
 
   @VisibleForTesting static final String HTTP_RPC_SYSTEM_NAME = "http";
+
   private static final String REDACTED_VALUE = "REDACTED";
+  // Required by OpenTelemetry semantic conventions:
+  //     https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/#url-full
   private static final Set<String> SENSITIVE_QUERY_KEYS =
       Collections.unmodifiableSet(
-          new HashSet<>(Arrays.asList("AWSAccessKeyId", "Signature", "sig", "X-Goog-Signature")));
+          new HashSet<>(
+              Arrays.asList(
+                  "AWSAccessKeyId",
+                  "Signature",
+                  "sig",
+                  "X-Goog-Signature",
+                  // Google uses this as a key in resumable uploads.
+                  "upload_id")));
 
   private final HttpRequestInitializer delegate;
   private final Tracer tracer;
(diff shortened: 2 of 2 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
```diff
@@ -0,0 +1,195 @@
...
+
+  private Span getSpan(String httpMethod, String url, String host, Integer port, Long requestBodySize) {
+    //TODO: Determine span name: {method} {url.template} or {method}
+    Span span =
+        BigQueryTelemetryTracer.newSpanBuilder(tracer, httpMethod)
+            // OpenTelemetry semantic convention attributes
+            .setAttribute(HTTP_REQUEST_METHOD, httpMethod)
+            .setAttribute(URL_FULL, url)
+            .setAttribute(BigQueryTelemetryTracer.SERVER_ADDRESS, host)
+            .setAttribute(URL_DOMAIN, BIGQUERY_DOMAIN)
```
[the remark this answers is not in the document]
ldetmer: its in the requirement, but changing to be the host

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
```diff
@@ -0,0 +1,286 @@
...
+      HttpResponse response, String httpMethod, Span span) {
+    String actualMethod = response.getRequest().getRequestMethod();
+    if (actualMethod != null && httpMethod == null) {
+      span.updateName(actualMethod);
+      span.setAttribute(HTTP_REQUEST_METHOD, actualMethod);
+    }
+  }
+
+  @VisibleForTesting
+  static String sanitizeUrlFull(String url) {
```
blakeli0: Is this needed? I know theoretically url can be anything, but when the url is passed from the client, it can only be of certain format. Maybe it is only a Bigquery thing because it is Apiary based? cc: @westarle Do we do this in other languages?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
```diff
@@ -0,0 +1,286 @@
...
+    // not available until after the http request execution
+    addRequestBodySizeToSpan(request, span);
+    checkForUpdatedRequestMethod(response, httpMethod, span);
+
+    addResponseBodySizeToSpan(response, span);
+    span.setAttribute(HTTP_RESPONSE_STATUS_CODE, statusCode);
+  }
+
+  private static void addExceptionToSpan(IOException e, Span span) {
+    span.recordException(e);
```
blakeli0: What exactly does `recordException` do? Does it automatically creates a few attributes? It seems like a good practice, but it is not listed as a requirement. So I want to make sure it does not conflict with existing attributes.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
```diff
@@ -0,0 +1,286 @@
...
+    String errorMessage = "HTTP " + statusCode;
+    try {
+      String statusMessage = response.getStatusMessage();
+      if (statusMessage != null && !statusMessage.isEmpty()) {
+        errorMessage = statusMessage;
+      }
+    } catch (Exception ex) {
+      // Ignore
+    }
+    span.setAttribute(BigQueryTelemetryTracer.STATUS_MESSAGE, errorMessage);
```
jinseopkim0: Are there any data privacy concerns with exporting the error message? I assume BigQuery error messages could potentially contain sensitive details like SQL query snippets, table names, or row data.

jinseopkim0 (changes requested):

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -107,10 +107,19 @@ public HttpBigQueryRpc(BigQueryOptions options) {
     HttpTransportOptions transportOptions = (HttpTransportOptions) options.getTransportOptions();
     HttpTransport transport = transportOptions.getHttpTransportFactory().create();
     HttpRequestInitializer initializer = transportOptions.getHttpRequestInitializer(options);
+
+    String resolvedBigQueryRootUrl = options.getResolvedApiaryHost("bigquery");
+
+    if (options.isOpenTelemetryTracingEnabled() && options.getOpenTelemetryTracer() != null) {
```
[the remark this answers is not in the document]
westarle: Our product decision today: we will preserve the existing spans and enablement and make non-breaking changes (adding attributes).

This continues to provide product-specific attributes and naming and avoids breaking BigQuery customers.

We considered creating a new enablement and forking spans but (1) the existing enablement in Beta allows for change and (2) it is very difficult to explain when a customer might want each type of CLIENT span.

In the future, we can decide whether it's worth it to adjust things like span name, structural hierarchy or enablement API by reviewing with the BQ team.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
westarle: recordException -- please skip it, we will use logs for this.

The span could have the exception type if it is convenient.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
```diff
@@ -0,0 +1,286 @@
...
+    Integer port = request.getUrl().getPort();
+
+    Span span = createHttpTraceSpan(httpMethod, url, host, port);
+
+    HttpResponseInterceptor originalInterceptor = request.getResponseInterceptor();
+    request.setResponseInterceptor(
+        response -> {
+          addCommonResponseAttributesToSpan(
+              request, response, span, httpMethod, response.getStatusCode());
+          span.setStatus(StatusCode.OK);
```
westarle, line 106: we should leave the status UNSET (it's weird, see https://github.com/googleapis/google-cloud-rust/issues/4938)

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpTracingRequestInitializer.java
westarle: Good thought!

if it's in OpenTelemetry semconv it should be OK.

Otherwise let's document it at the enablement site.

see go/client-libraries:o11y-controls

